PR #6, opened 2023-03-14
FW-4119 Add base model class

NolanVH:
This change adds the base model class which can be inherited by other models. 
The base class contains the following fields:

- id - a UUID
- is_trashed - a field to indicate if the model has been deleted
- created_by - a foreign key to a user
- created - date and time which is automatically updated on creation using the auto_now_add option
- last_modified_by - a foreign key to a user
- last_modified - date and time which is automatically updated when the model is modified using the auto_now option

The date fields are automatically handled and updated but the user fields must be updated when the documents are created or modified.

### DIFF
--- a/fv_be/app/models/base.py
+++ b/fv_be/app/models/base.py
@@ -30,11 +30,11 @@ class BaseModel(RulesModel):
     created_by = models.ForeignKey(
         User, on_delete=models.SET_NULL, null=True, default=None
     )
-    created_by_date = models.DateTimeField(auto_now_add=True)
+    created = models.DateTimeField(auto_now_add=True)
     last_modified_by = models.ForeignKey(
         User, on_delete=models.SET_NULL, null=True, default=None
     )
-    last_modified_date = models.DateTimeField(auto_now=True)
+    last_modified = models.DateTimeField(auto_now=True)
 
     @property
     def created_by(self):
--- a/fv_be/app/models/base.py
+++ b/fv_be/app/models/base.py
@@ -1,0 +1,45 @@
+import uuid
+
+# this will be replaced by our custom User class when we create the site models
+from django.contrib.auth.models import User
+from django.db import models
+from rules.contrib.models import RulesModel
+
+
+class BaseModel(RulesModel):
+    """
+    Base model for all FirstVoices Backend models, with standard fields and support for rules-based permissions.
+
+    Date fields values are generated automatically, but user fields (created_by and last_modified_by) are required when
+    creating new data.
+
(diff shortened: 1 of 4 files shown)


sahil97 (approved):

bronson-c (approved):
LGTM 👍

fv_be/app/models/base.py
```diff
@@ -0,0 +1,45 @@
...
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    is_trashed = models.BooleanField(default=False)
+    created_by = models.ForeignKey(
+        User, on_delete=models.SET_NULL, null=True, default=None
+    )
+    created_by_date = models.DateTimeField(auto_now_add=True)
+    last_modified_by = models.ForeignKey(
+        User, on_delete=models.SET_NULL, null=True, default=None
+    )
+    last_modified_date = models.DateTimeField(auto_now=True)
```
gmcauliffe, line 37: Is there added value in the explicitness of suffix `_by_date`?
Could we simplify these two from `created_by_date` and `last_modified_by_date`
to `created` and `last_modified` ?

fv_be/app/models/base.py
```diff
@@ -0,0 +1,45 @@
...
+
+# this will be replaced by our custom User class when we create the site models
+from django.contrib.auth.models import User
+from django.db import models
+from rules.contrib.models import RulesModel
+
+
+class BaseModel(RulesModel):
+    """
+    Base model for all FirstVoices Backend models, with standard fields and support for rules-based permissions.
```
gmcauliffe, line 11: Very helpful level of detail in these comments 👍

fv_be/app/models/base.py
sarahfirstvoices: I like the idea of the shorter names

NolanVH: I agree that shorting naming could be used here. I have updated the naming in a small commit. Thanks for the suggestion @gmcauliffe.

caforbes (approved):
Looking good! :)

fv_be/app/models/base.py
```diff
@@ -0,0 +1,44 @@
...
+            "add": rules.is_staff,
+            "view": rules.is_authenticated,
+        }
+    See the django-rules docs [https://github.com/dfunckt/django-rules#permissions-in-models] for more details.
+    """
+
+    class Meta:
+        abstract = True
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
```
plasticviking, line 27: Is there a reason to prefer UUID rather than a bigint here?

PR #7, opened 2023-03-15
FW-4120 Add models for language family, sites, membership, site info,…

sarahfirstvoices:
… and base models for site content, and the corresponding admin forms.

Still to come in other tickets: a user model (pending Rob's existing auth work), and feature flags for sites.


fv_be/app/models/sites.py
```diff
@@ -0,0 +1,133 @@
...
+    # fvdialect:about_our_language
+    about_our_language = models.TextField(blank=True)
+
+    # fvdialect:about_us
+    about_us = models.TextField(blank=True)
+
+    # dc:description
+    description = models.TextField(blank=True)
+
+    # todo: discuss how to handle existing contributors which are stored as usernames / from dc:contributors
```
sarahfirstvoices, line 109: @gmcauliffe and @caforbes The Django way of finding out who has ever contributed to a site would be via the change log history, but I don't relish the idea of trying to add historical change logs as part of the data migration. Is there value in keeping this list of who has contributed to the nuxeo sites at least as a text field or something, or could we ignore it?

fv_be/app/models/sites.py
```diff
@@ -0,0 +1,133 @@
...
+    # todo: add featured words when words are available / from fvdialect:featured_words
+    # todo: add featured audio when media is available / from fvdialect:featured_audio
+    # todo: add keyboards when available / from fvdialect:keyboards
+    # todo: decide whether to import fvl:import_id (from the fvlegacy schema)
```
sarahfirstvoices, line 113: @gmcauliffe and @caforbes Do you think we need to migrate this legacy id field?

fv_be/app/models/sites.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def __str__(self):
+        # todo: i18n
+        return f"Site information for {self.site.title}"
+
+
+class Membership(BaseSiteContentModel):
+    """Represents a user's membership to a language site"""
+
+    # todo: mappings from nuxeo
```
sarahfirstvoices, line 124: I will track these down but didn't want to hold up the PR over it.

fv_be/app/models/sites.py
```diff
@@ -0,0 +1,133 @@
+from django.core.validators import validate_slug
+from django.db import models
+
+from fv_be.app.models import BaseModel
+from fv_be.users.models import User
+
+from .constants import Role, Visibility
+
+
+class LanguageFamily(BaseModel):
```
sarahfirstvoices, line 10: @caforbes Is this a good name for this model? It is pulling from the "FVLanguage" documents in nuxeo, but thought we might want to keep "Language" available as a name for future use. This is the model for the group headings that are shown on the Explore Languages page.

fv_be/app/models/base.py
caforbes, line 1: love these comments for migration!

fv_be/app/models/sites.py
caforbes: No, this is very confusing to me because "language family" is a different level of grouping which I would expect to map to FVLanguageFamily. (See the FPCC language status report which is grouped by family < language).

I would stick with Language here and replace all references to "language family" with just "language". I don't think anything would replace this core concept. (Future-feature-wise, sites with multiple "languages" will likely not be cross-"Language" in this sense, but cross-community or cross-Site).

We could discuss storing the language family information too, since this is known information. This is what the nuxeo directory structure was meant to encode -- it would be better in a 1:M relation with Language. However, I don't think we currently leverage this info in any feature, so I wouldn't prioritize it.

fv_be/app/models/sites.py
```diff
@@ -0,0 +1,133 @@
...
+    visibility = models.IntegerField(
+        choices=Visibility.choices, default=Visibility.TEAM
+    )
+
+
+class SiteInformation(BaseSiteContentModel):
+    """Contains the access-controlled information about a language site."""
+
+    # details come from fvdialect document
+
```
caforbes, line 90: AFAIK the following fields are all related to the V1 "portal" and are not leveraged in V2:
- fvdialect:contact_information
- fvdialect:greeting
- fvdialect:about_our_language
- fvdialect:about_us
- dc:description
- fvdialect:featured_words
- fvdialect:featured_audio
- fvdialect:keyboards
In V2 these are handled with widgets, so rather than migrate and store them here, I would suggest migrating them as Widgets in whichever widget is the best fit, and dropping them otherwise. -- @gmcauliffe ?

fv_be/app/models/sites.py
caforbes: I think it might be fine to just store the lastContributor as last_modified_by, and not the entire history?

fv_be/app/models/sites.py
caforbes: I think we could leave it out, and look back at the archived Nuxeo data if we need to reference it ever.

fv_be/app/models/sites.py
```diff
@@ -0,0 +1,133 @@
...
+    def __str__(self):
+        # todo: i18n
+        return f"Site information for {self.site.title}"
+
+
+class Membership(BaseSiteContentModel):
+    """Represents a user's membership to a language site"""
+
+    # todo: mappings from nuxeo
+    user = models.ForeignKey(User, on_delete=models.CASCADE, db_index=True)
```
sahil97, line 125: a db_index is automatically created for Foreign Keys, should we still explicitly set it here ?

fv_be/app/models/base.py
```diff
@@ -24,21 +24,36 @@ class Meta:
...
     created = models.DateTimeField(auto_now_add=True)
+
+    # from dc:modified
     last_modified_by = models.ForeignKey(
         settings.AUTH_USER_MODEL,
         on_delete=models.SET_NULL,
         null=True,
         default=None,
         related_name="modified_%(app_label)s_%(class)s",
+        db_index=True,
```
sarahfirstvoices, line 55: move db indexes to the date fields

fv_be/app/models/sites.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # from fvlanguage type
+
+    class Meta:
+        # todo: i18n
+        verbose_name = "language family"
+        verbose_name_plural = "language families"
+
+    # from dc:title
+    title = models.CharField(max_length=200)
```
sarahfirstvoices, line 23: Fields for consideration: bcp code, language family.

sarahfirstvoices: More fields: maybe colour (possibly by language family), name variations.

fv_be/app/models/sites.py
```diff
@@ -0,0 +1,133 @@
...
+    # from fva:language
+    language_family = models.ForeignKey(LanguageFamily, on_delete=models.DO_NOTHING)
+
+    # from state (will have to be translated from existing states to new visibilities)
+    visibility = models.IntegerField(
+        choices=Visibility.choices, default=Visibility.TEAM, db_index=True
+    )
+
+    # todo: add logo field when media is ready / from fvdialect:logo
+    # todo: add banner field when media is ready / from fvdialect:background_top_image
```
sarahfirstvoices, line 54: Suggestion to move banner to the Homepage

fv_be/app/models/sites.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def __str__(self):
```
sarahfirstvoices, line 56: Check the v2 sites endpoints for any extra fields

fv_be/app/models/sites.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # todo: discuss how to handle existing contributors which are stored as usernames / from dc:contributors
+    # todo: add featured words when words are available / from fvdialect:featured_words
+    # todo: add featured audio when media is available / from fvdialect:featured_audio
+    # todo: add keyboards when available / from fvdialect:keyboards
+    # todo: decide whether to import fvl:import_id (from the fvlegacy schema)
+    # todo: as part of pages and widgets epic / from widgets:active
+
+    def __str__(self):
+        # todo: i18n
```
sarahfirstvoices, line 117: We can move site_menu to the Site for now. Log a ticket to consider splitting it into its own model for easier editing.

fv_be/app/models/sites.py
sarahfirstvoices: i'll remove all the foreign key ones. thanks

config/settings/base.py
```diff
@@ -85,6 +84,7 @@
 
 LOCAL_APPS = [
     "fv_be.users",
+    "fv_be.app",
```
bronson-c, line 87: Small change to base configuration to fix errors when unregistering models on the admin site

sahil97 (approved):
LGTM

fv_be/app/migrations/0001_initial.py
bronson-c, line 1: Does this auto generate file need to be present? I noticed while starting the site another file `0002_alter_membership_role_alter_site_visibility.py` is generated but it is not included on this PR.

bronson-c (approved):
I just have the one question about the file in the migrations folder that I was curious about, LGTM 👍

fv_be/app/models/sites.py
```diff
@@ -0,0 +1,150 @@
...
+
+    # from dc:title
+    alternate_names = models.TextField(max_length=200, blank=True)
+
+    # from fva:family
+    language_family = models.ForeignKey(LanguageFamily, on_delete=models.PROTECT)
+
+    # from fvdialect:bcp_47
+    # BCP 47 Spec: https://www.ietf.org/rfc/bcp/bcp47.txt
+    language_code = models.TextField(max_length=100, blank=True)
```
caforbes, line 55: I think 20 would be a better (very generous) max. These are usually short, e.g. `en`, `haa`, `kwk-napa`. And maybe CharField rather than TextField.

fv_be/app/migrations/0001_initial.py
NolanVH: Committing migrations is generally a good thing to do once we are running in production, but if it is becoming a problem for the development stage we could add the directory to the .gitignore file for now. You are probably generating the ```0002``` file because you may have an existing local ```0001``` file which is slightly different than the committed one. If you delete the migrations and pull a fresh copy of the branch you should get no changes detected.

sarahfirstvoices: I do think we should keep the migrations in version control, because they will be required for ci/cd to work properly. Eventually we will want to modify our deployed db rather than installing it from scratch, and migration files are the way to do that. I'll check about that file-- maybe we missed one!

bronson-c: Just tested again with a fresh branch, the issue of the 2nd file is resolved 👍

fv_be/app/models/sites.py
sarahfirstvoices: Thanks! Changed it.

fv_be/app/migrations/0001_initial.py
sarahfirstvoices: We checked this offline and the current file is correct. I think maybe we just had a merge-related mixup. Good to check tho

NolanVH (approved):

PR #12, opened 2023-03-16
Fw 4053 create dictionary data models

sahil97:


fv_be/app/models/dictionary.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def __str__(self):
+        return _("Translation in %(language)s: %(translation)s.") % {
+            "language": self.language,
+            "translation": self.translation,
+        }
+
+
+class AlternateSpelling(BaseModel):
+    """Model for alternate spellings associated to each dictionary entry."""
```
caforbes, line 59: Note: I don't think Alternate Spellings are set up on our V2 edit forms right now. We can still include them, but it will be a later feature to add them to the edit forms. (It would also be awesome to make text in this field part of the site search at that time.)

fv_be/app/models/dictionary.py
```diff
@@ -0,0 +1,125 @@
...
+    type = models.CharField(
+        max_length=6, choices=TYPE_OF_ENTRY_CHOICES, default=WORD_ENUM_KEY
+    )
+    # todo: Link to site model when available
+    # may be inherited from an abstract base class or mixin later
+    # site = models.ForeignKey(
+    #     'Site',
+    #     on_delete=models.CASCADE
+    # )
+    notes = models.ForeignKey("Note", on_delete=models.SET_NULL, null=True)
```
caforbes, line 104: The V2 UI currently supports 1 Entry with multiple Notes, so it should be Notes that have the ForeignKey to the entry. I see that the dictionary model draft suggests M:M for this relation but I don't think this makes sense. Even if multiple entries have the same Note text, the actual Note should not be shared across different entries -- a user wouldn't expect editing the Note on one entry to affect the Note on another entry. -- Happy to discuss this more if needed.

fv_be/app/models/dictionary.py
```diff
@@ -0,0 +1,125 @@
...
+    acknowledgments = models.ForeignKey(
```
caforbes, line 105: Same issue as with Notes.

fv_be/app/models/part_of_speech.py
```diff
@@ -0,0 +1,24 @@
...
+class PartOfSpeech(BaseModel):
+    """Model for Categories."""
+
+    # Fields
+    # todo: name for the following field, can we put title ? or will it be very similar to label ?
+    # since other models have a title field
+    name = models.CharField(
+        max_length=200
+    )  # from wiki: name of the part of speech in snake case e.g.intransitive_verb
+    label = models.CharField(max_length=200)
```
caforbes, line 16: Do they need to have both a name and a label? Seems like we could use an ID from the base model as the primary key, ditch the snake_case label from nuxeo, and just keep the formatted one as a title.

fv_be/app/models/part_of_speech.py
```diff
@@ -0,0 +1,24 @@
...
+    # todo: name for the following field
+    parent = models.ManyToManyField("self", blank=True)
```
caforbes, line 18: Same as for Categories, parts of speech can only have 1 parent.

fv_be/app/models/part_of_speech.py
```diff
@@ -0,0 +1,24 @@
+from django.db import models
+
+# FirstVoices
+from .base import BaseModel
+
+
+class PartOfSpeech(BaseModel):
+    """Model for Categories."""
```
caforbes, line 8: Text here and in Meta should be "PartsOfSpeech"

caforbes (changes requested):

fv_be/app/models/category.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Category(BaseModel):
+    """Model for Categories."""
+
+    # Fields
+    title = models.CharField(max_length=200)
+    description = models.TextField()
+    # todo: Add a nesting check for maximum one level of nesting
+    # i.e. Parent Categories can have children but those children cannot have more children
+    parent = models.ManyToManyField("self", blank=True)
```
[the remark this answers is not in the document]
sarahfirstvoices: I think this also needs symmetrical=False so we can tell which way the parent relationship points

fv_be/app/models/dictionary.py
```diff
@@ -0,0 +1,125 @@
...
+
+    # Fields
+    translation = models.CharField(max_length=200)
+    language = models.CharField(
+        max_length=2, choices=LANGUAGE_CHOICES, default=ENGLISH_ENUM_KEY
+    )
+    # from fv-word:part_of_speech
+    part_of_speech = models.ManyToManyField(PartOfSpeech)
+    # todo: more representative name for the following attribute ?
+    parent = models.ForeignKey("DictionaryEntry", on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
sahil97: Yes, this is a better idea I believe.

fv_be/app/models/dictionary.py
sahil97: Okay, I believe we could remove it for now and then add it back later when we're adding stuff to it.

caforbes: There is existing data in the fields though, so does that mean we should add it now since it will have to be part of migration? Or is the idea that we would get back to this and add it before we actually migrate any data? (@sarahfirstvoices ?)

fv_be/app/models/dictionary.py
```diff
@@ -0,0 +1,97 @@
...
+    # from fv:custom_order
+    custom_order = models.CharField(max_length=200, blank=True)
+    # from fv-word:available_in_games, fvaudience:games
+    exclude_from_games = models.BooleanField(default=False)
+    # from fvaudience:children fv:available_in_childrens_archive
+    # exclude_from_kids can be a shared mixin for dictionary_entries, songs, stories and media
+    exclude_from_kids = models.BooleanField(default=False)
+    batch_id = models.CharField(max_length=255, blank=True)
+    # from fv:related_assets, fv-word:related_phrases
+    related_dictionary_entries = models.ManyToManyField("self", blank=True)
```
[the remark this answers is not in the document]
sahil97: Yes, they are not symmetrical in Nuxeo which looked a bit non-intuitive to me. e.g. If _Word1_ has _Word2_ in its related words, _Word2_ should also display _Word1_ in its related words. I believe we should add the symmetrical flag.
We can have more discussion about the expected behaviour here.

fv_be/app/models/dictionary.py
sahil97: Oh okay, If there is data in the fields, I would suggest we should add the model now and then add that data as a part of the migration, even if it is not setup on the edit forms.

fv_be/app/models/part_of_speech.py
sahil97: I would agree with you @caforbes. And maybe when we only have one field, we can name it 'title' to make it consistent with the rest of the models ?
Tagging @gmcauliffe for his view about doing away with nuxeo snake_case lables.

fv_be/app/models/dictionary.py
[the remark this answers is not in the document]
sahil97: Great suggestion. I'll add that. Any extra fields you suggest we should add now ? or we can leave it for later as well when the need arises.

fv_be/app/models/part_of_speech.py
sarahfirstvoices: Just remember that it will go through i18n before being displayed, so the db value doesn't need to be especially well formatted

fv_be/app/models/dictionary.py
```diff
@@ -0,0 +1,153 @@
...
+    text = models.TextField()
+    dictionary_entry = models.ForeignKey(
+        "DictionaryEntry", on_delete=models.CASCADE, related_name="notes"
+    )
+
+    def __str__(self):
+        return self.text
+
+
+class Acknowledgement(BaseModel):
```
bronson-c, line 23: Same comment as the Notes model applies here. I'm ok with keeping it as-is, just thought I'd mention it.

fv_be/app/models/dictionary.py
```diff
@@ -0,0 +1,153 @@
...
+
+    class Meta:
+        verbose_name_plural = "DictionaryEntries"
+
+    def __str__(self):
+        return self.title
+
+
+class DictionaryEntrySelfAssociation(models.Model):
+    from_dictionary_entry = models.ForeignKey(DictionaryEntry, on_delete=models.CASCADE)
```
bronson-c, line 148: Should we cascade on delete here? I'd assume if a related entry is deleted that we would just null the relation.

fv_be/app/models/part_of_speech.py
```diff
@@ -0,0 +1,23 @@
...
```
bronson-c, line 8: Should say "Parts of Speech"

fv_be/app/models/part_of_speech.py
```diff
@@ -0,0 +1,23 @@
+from django.db import models
+
+# FirstVoices
+from .base import BaseModel
+
+
+class PartOfSpeech(BaseModel):
```
bronson-c, line 7: Would there be any reason for the parts of speech model to have a foreign key linking to the site the category belongs to?

fv_be/app/models/part_of_speech.py
```diff
@@ -0,0 +1,23 @@
...
+    """Model for Categories."""
+
+    # Fields
+    # todo: name for the following field, can we put title ? or will it be very similar to label ?
+    # since other models have a title field
+    name = models.CharField(max_length=200)
```
bronson-c, line 13: I would rename to title for consistency (if we are keeping this field)

fv_be/app/models/part_of_speech.py
sahil97: Parts of speech will be an independent list as of now, so it will not require a foreign key to the site as of now.
Future plans are for each site to maintain their own list for parts of speech but they are not implemented as of now.

fv_be/app/admin/admin.py
```diff
@@ -5,11 +5,30 @@
...
+    Pronunciation,
+    Translation,
+)
+from fv_be.app.models.part_of_speech import PartOfSpeech
 from fv_be.app.models.sites import Site
 
 from .base_admin import BaseAdmin
 from .sites_admin import MembershipInline
 
+dictionary_models = [
```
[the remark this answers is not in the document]
sahil97: There is no specific purpose, I was just trying to write more concise code, i.e. just registering this list instead of registering 6 models separately coming from the same file. I believe it may be similar in your case of character models as well.

fv_be/app/models/dictionary.py
```diff
@@ -0,0 +1,152 @@
+from django.db import models
+from django.utils.translation import gettext as _
+
+# FirstVoices
+from .base import BaseModel
+from .category import Category
+from .part_of_speech import PartOfSpeech
+
+
+class Note(BaseModel):
```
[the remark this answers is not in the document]
sahil97: Good point. If the model can be referred in other models, provided their notes field follow the same structure, to avoid duplication, we can move this out Dictionary file and make it a general table which can be referred to by any site. Or if the structure/field for different models is different but similar name, I agree with your point of renaming it to DictionaryNote.
Tagging @gmcauliffe @sarahfirstvoices @caforbes for their take on this.

fv_be/app/models/dictionary.py
gmcauliffe: Agreed. Good catch :)

fv_be/app/models/dictionary.py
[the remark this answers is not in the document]
gmcauliffe: See below.

fv_be/app/models/dictionary.py
sahil97: Yes. We would not want the related entry to be delete in case one entry is deleted. I'll try to explain my understanding of this but please correct me if I'm wrong.

In this case, this DictionaryEntrySelfAssociation model is an intermediary ("through" in django terms) table to store the many to many relation. So what the CASCADE implies is, if word1 and word2 are related, and one word gets deleted, this just deletes the relation b/w those words, since there is no point in keeping a relation which has null in either 'from' or 'to' in the association table.

fv_be/app/models/category.py
```diff
@@ -0,0 +1,37 @@
...
+from .sites import BaseSiteContentModel
+
+
+class Category(BaseSiteContentModel):
+    """Model for Categories."""
+
+    # Fields
+    title = models.CharField(max_length=200)
+    description = models.TextField()
+    # i.e. Parent Categories can have children but those children cannot have more children
```
sarahfirstvoices, line 15: I think it is good to have this comment here, but it is a bit confusing to me because it refers to having children when the field is actually for specifying a parent. Could we make the comment align? Something like "A category may have a parent, but grandparents are not allowed and this is enforced in the clean method"

fv_be/app/models/category.py
```diff
@@ -0,0 +1,37 @@
...
+    parent = models.ForeignKey("self", blank=True, null=True, on_delete=models.PROTECT)
+
+    class Meta:
+        verbose_name_plural = _("Categories")
```
sarahfirstvoices, line 19: Probably worth including the verbose_name as well, for the i18n

fv_be/app/models/category.py
```diff
@@ -0,0 +1,37 @@
...
+    def __str__(self):
+        return self.title
+
+    def clean(self):
+        self.is_cleaned = True
+        # Enforce only one max level of nesting
+        parent_category = self.parent
+        if parent_category and parent_category.parent:
+            raise ValidationError(
+                _("Choosing categories that are themselves children of other categories is not allowed.")
```
sarahfirstvoices, line 30: Similar to above, could we keep to the "parent" language in this message? Something like "Parent category cannot have a parent."

fv_be/app/models/dictionary.py
```diff
@@ -0,0 +1,154 @@
...
+from .sites import BaseSiteContentModel, BaseControlledSiteContentModel
+from .category import Category
+from .part_of_speech import PartOfSpeech
+
+
+class Note(BaseSiteContentModel):
+    """Model for notes associated to each dictionary entry."""
+
+    # from fv:notes,fv:general_note, fv:cultural_note, fv:literal_translation, fv-word:notes, fv-phrase:notes
+    text = models.TextField()
```
sarahfirstvoices, line 14: Sorry if this was already discussed, but do we need large text notes? If there is a max length currently enforced in the ui, maybe we could stick to that

fv_be/app/models/dictionary.py
```diff
@@ -0,0 +1,154 @@
...
+        return self.text
+
+
+class Acknowledgement(BaseSiteContentModel):
+    """Model for acknowledgments associated to each dictionary entry."""
+
+    # from fv:acknowledgments, fv:source, fv:reference, fv-word:acknowledgement, fv-phrase:acknowledgement
+    text = models.TextField()
+    # todo: Confirm if this should be moved to dictionaryEntry, M:M relation
+    dictionary_entry = models.ForeignKey(
```
sarahfirstvoices, line 29: I think this todo can be removed. ForeignKey looks correct to me

fv_be/app/models/dictionary.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def __str__(self):
+        return self.text
+
+
+class Acknowledgement(BaseSiteContentModel):
+    """Model for acknowledgments associated to each dictionary entry."""
+
+    # from fv:acknowledgments, fv:source, fv:reference, fv-word:acknowledgement, fv-phrase:acknowledgement
+    text = models.TextField()
```
sarahfirstvoices, line 27: Similar to the Notes-- does this need to be a TextField or could it be a CharField with a maximum length?

fv_be/app/models/dictionary.py
```diff
@@ -0,0 +1,154 @@
...
+    # Fields
+    # from dc:title, relatively more max_length due to phrases
+    # todo: max_length may be modified after doing some analysis on the length of current phrases
+    title = models.CharField(max_length=800)
+    type = models.CharField(
+        max_length=6,
+        choices=TypeOfDictionaryEntry.choices,
+        default=TypeOfDictionaryEntry.WORD,
+    )
+    # todo: Link to site model when available
```
sarahfirstvoices, line 110: This todo can be removed now that you are using the base model

fv_be/app/models/dictionary.py
```diff
@@ -0,0 +1,154 @@
...
+    # from fv-word:categories, fv-phrase:phrase_books
+    categories = models.ForeignKey(
+        Category,
+        blank=True,
+        null=True,
+        on_delete=models.SET_NULL,
+        related_name="dictionary_entries",
+    )
+    # from fv:custom_order
+    custom_order = models.CharField(max_length=200, blank=True)
```
sarahfirstvoices, line 125: I think this will need to have the same length as the title, right? @caforbes @bronson-c

fv_be/app/models/dictionary.py
```diff
@@ -0,0 +1,154 @@
...
+    # from fv-word:available_in_games, fvaudience:games
+    exclude_from_games = models.BooleanField(default=False)
+    # from fvaudience:children fv:available_in_childrens_archive
+    # exclude_from_kids can be a shared mixin for dictionary_entries, songs, stories and media
+    exclude_from_kids = models.BooleanField(default=False)
+    batch_id = models.CharField(max_length=255, blank=True)
```
sarahfirstvoices, line 131: Does batch_id need a migration comment?

fv_be/app/models/dictionary.py
```diff
@@ -0,0 +1,154 @@
...
+    )
+
+    class Meta:
+        verbose_name_plural = _("DictionaryEntries")
+
+    def __str__(self):
+        return self.title
+
+
+class DictionaryEntrySelfAssociation(models.Model):
```
sarahfirstvoices, line 148: I think this name could maybe be simpler... Maybe something like RelatedDictionaryEntry or DictionaryEntryLink?

fv_be/app/models/part_of_speech.py
```diff
@@ -0,0 +1,19 @@
...
+# FirstVoices
+from .base import BaseModel
+
+
+class PartOfSpeech(BaseModel):
+    """Model for Parts Of Speech."""
+
+    # Fields
+    title = models.CharField(max_length=200)
+    parent = models.ForeignKey("self", null=True, blank=True, on_delete=models.PROTECT)
```
sarahfirstvoices, line 13: Please add a related_name here, maybe 'children'

fv_be/app/models/dictionary.py
gmcauliffe: @sarahfirstvoices Wouldn't it be better to make Acknowledgement a M:M relationship to avoid duplication for those teams that want to add the same acknowledgement to multiple entries?

caforbes: By pointing multiple entries to the same acknowledgement, we would introduce a behaviour where editing the acknowledgement on one entry would edit it for all other entries that have the same acknowledgement. So if you had an acknowledgement for "page 5" on ten different entries, and then you edited one to "page 6", they would all update to "page 6". I don't think this is what we want for our use case, I don't think it would be intuitive for the user. Same issue for notes.

fv_be/app/models/dictionary.py
caforbes: It will not be the same length as the title, it could be either shorter or longer.

caforbes: Oh but oops the point is that it should be way closer to the current max value of the title -- probably the same would be safe.

fv_be/app/models/dictionary.py
gmcauliffe: @caforbes If we added M:M on the acknowledgement model we wouldn't necessarily have to make use of the ability to make multiple references to all acknowledgements, but by putting this in the model we could use it for specific cases where a team wants to add the same acknowledgment to multiple entries. We already have teams who have added the same acknowledgement to whole batches, and if they ever wanted to modify or update that acknowledgment it would need to be modified on each and every one.

fv_be/app/admin/admin.py
```diff
@@ -0,0 +1,64 @@
...
+admin.site.unregister(SocialAccount)
+admin.site.unregister(SocialApp)
+admin.site.unregister(SocialToken)
+
+# Dictionary models
+
+# todo: make suitable admin objects to register for each model
+# todo: make custom admin forms to prevent self selection in ManyToMany fields referring to self
+# ref: https://stackoverflow.com/questions/869856/
+admin.site.register(dictionary_models)
```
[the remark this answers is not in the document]
sahil97: Yes, that is correct. I'm working on that in a separate branch for ticket FW-4184. This code is only here for testing purposes. Will be removed in the upcoming PR.

fv_be/app/models/part_of_speech.py
```diff
@@ -0,0 +1,20 @@
...
+# FirstVoices
+from .base import BaseModel
+
+
+class PartOfSpeech(BaseModel):
+    """Model for Parts Of Speech."""
+
+    # Fields
+    title = models.CharField(max_length=200)
+    parent = models.ForeignKey("self", null=True, blank=True, on_delete=models.PROTECT, related_name="children")
```
caforbes, line 13: Parts of speech should have the same "no grandparents" restriction as categories do.

fv_be/app/models/dictionary.py
sahil97: Max length can be added after research on basis of ticket FW-4196. Marking it resolved for now.

fv_be/app/models/dictionary.py
sahil97: Current resolution: Only foreignKey relationship for acknowledgements for now. Moreover, prefix to be added to dictionary models to be specific. If the models need to be reused, we can pull them out of DictionaryModel file and make it a separate table.

fv_be/app/models/dictionary.py
sahil97: Max length can be added after research on basis of ticket FW-4196. Marking it resolved for now.

sarahfirstvoices (approved):

PR #16, opened 2023-03-21
FW-4177 Add Site Menu Model

NolanVH:
This change adds the site menu model and a model to hold the default values for a menu model. The purpose of this separate model which contains the default value is to make the default menus across multiple sites editable without code changes.

Additionally, this change adds search filtering to admin model views.


fv_be/app/models/sites.py
```diff
@@ -164,3 +164,21 @@ class Meta:
 
     def __str__(self):
         return f"{self.key}: {self.is_enabled} ({self.site})"
+
```
sarahfirstvoices, line 167: The idea here is to only store a SiteMenu if a site has customized it. Otherwise we can return a default value that is stored in the AppJson model.

bronson-c (approved):
Looks good to me, just need to fix merge conflicts with dev 👍

caforbes (approved):

PR #17, opened 2023-03-23
FW-4123: enable related dictionary entries, fix dependency

bronson-c:
This PR enables the related_dictionary_entries field on the character models, and includes the appropriate admin changes. It also fixes a circular dependency error present in dev.


fv_be/app/models/characters.py
```diff
@@ -30,9 +31,13 @@ class Meta:
...
     # from fvcharacter:related_words
-    # TODO: add this as a foreign key to the dictionary entry model once it's created
-    # related_dictionary_entries = models.ForeignKey(
-    #     DictionaryEntry, on_delete=models.SET_NULL, blank=True, null=True, related_name="characters")
+    related_dictionary_entries = models.ForeignKey(
+        DictionaryEntry,
+        on_delete=models.SET_NULL,
+        blank=True,
+        null=True,
+        related_name="dictionary_entries",
```
sahil97, line 39: I believe the related name here would be something on the lines of "characters" or such. As it is what we would access from a dictionary entry with syntax such as `DictionaryEntry.characters.all()` to get the characters linked to a DictionaryEntry.

bronson-c: Nice catch

fv_be/app/models/characters.py
```diff
@@ -30,9 +31,13 @@ class Meta:
     notes = models.TextField(blank=True)
 
     # from fvcharacter:related_words
-    # TODO: add this as a foreign key to the dictionary entry model once it's created
-    # related_dictionary_entries = models.ForeignKey(
-    #     DictionaryEntry, on_delete=models.SET_NULL, blank=True, null=True, related_name="characters")
+    related_dictionary_entries = models.ForeignKey(
```
sarahfirstvoices, line 34: I think this actually needs to be a many-to-many relation, right?

sarahfirstvoices (changes requested):

fv_be/app/models/characters.py
bronson-c: Changed from an FK to a many to many field, I do not think we need a "through" class for this relation but let me know if we do.

fv_be/app/models/characters.py
```diff
@@ -30,9 +31,13 @@ class Meta:
     notes = models.TextField(blank=True)
 
     # from fvcharacter:related_words
-    # TODO: add this as a foreign key to the dictionary entry model once it's created
-    # related_dictionary_entries = models.ForeignKey(
-    #     DictionaryEntry, on_delete=models.SET_NULL, blank=True, null=True, related_name="characters")
+    related_dictionary_entries = models.ManyToManyField(
```
bronson-c, line 34: Found a bug with this implementation, do not merge for now

fv_be/app/admin/characters_admin.py
```diff
@@ -46,6 +48,7 @@ class CharacterInline(BaseInlineAdmin):
         "sort_order",
         "approximate_form",
         "notes",
+        "related_dictionary_entries",
     ) + BaseInlineAdmin.fields
     readonly_fields = BaseInlineAdmin.readonly_fields + CharacterAdmin.readonly_fields
```
NolanVH, line 54: Just a suggestion but the default admin display for ManyToMany fields is somewhat unintuitive and it may be useful to add `filter_horizontal = ("related_dictionary_entries",)` here. It might also be worth adding this to the CharacterInline as well.

fv_be/app/admin/characters_admin.py
```diff
@@ -13,11 +13,13 @@ class CharacterAdmin(BaseSiteContentAdmin):
         "approximate_form",
         "site",
         "notes",
+        "related_dictionary_entries",
     )
     list_display = (
         "title",
         "sort_order",
         "approximate_form",
+        "related_dictionary_entries",
```
NolanVH, line 22: I believe for a ManyToMany field the list_display will give an error. There are workarounds for adding a ManyToMany field to the list view but it may be better to leave that field as only editable from the detail view.

NolanVH (commented):
After the bug you found with the ManyToMany field is fixed these changes all look good to me. Thanks @bronson-c!

fv_be/app/admin/characters_admin.py
bronson-c: I just added that it's a much better UI. Thanks!

fv_be/app/models/characters.py
bronson-c: After giving it a shot I believe that this implementation with the horizontal filter is simpler. Is there a specifc reason to specify a through model?

sarahfirstvoices: It's one of those things that is helpful for the future-- it gives us an explicit db table name and helps decouple the database schema from django's internal implementation choices. Here's a link with a longer explanation: https://medium.com/@DoorDash/tips-for-building-high-quality-django-apps-at-scale-a5a25917b2b5#6eea

bronson-c: Ah I see, ok I will try to get my through implementation running the same as the current one

PR #18, opened 2023-03-23
Fw 4184 add suitable admin models for all dictionary models

sahil97:


fv_be/app/models/dictionary.py
```diff
@@ -2,39 +2,50 @@
...
 from .sites import BaseSiteContentModel, BaseControlledSiteContentModel
 from .category import Category
 from .part_of_speech import PartOfSpeech
 
 
-class DictionaryNote(BaseSiteContentModel):
+class BaseDictionaryContentModel(BaseModel):
+    """
+    Base model for Dictionary models which all have DictionaryEntry as a foreign key and
+    require site as a property but not as a field.
```
sarahfirstvoices, line 14: Minor comment edit: these "require" a DictionaryEntry and "have" site as a property.

fv_be/app/admin/base_admin.py
```diff
@@ -70,16 +71,30 @@ def item_id(self, instance):
...
     def save_model(self, request, obj, form, change):
         if not change:
             obj.created_by = request.user
 
         obj.last_modified_by = request.user
         super().save_model(request, obj, form, change)
+
+
+# todo: better name for following class
+class AdminHideUtility(BaseAdmin):
```
sarahfirstvoices, line 94: Let's call this HiddenBaseAdmin or something similar, to make it clear that it is a base admin class.

fv_be/app/admin/base_admin.py
```diff
@@ -70,16 +71,30 @@ def item_id(self, instance):
...
+    """
+    This utility class lets the admin classes to be registered on the top level of admin interface, to get the reverse
+    links working for inline admin classes and also prevents the classes that inherit this to be shown at the top level.
+    """
```
sarahfirstvoices, line 98: This is a bit of a surprising implementation-- can you please add a comment with a link to a url that explains how this works?

fv_be/app/admin/base_admin.py
```diff
@@ -70,16 +71,30 @@ def item_id(self, instance):
...
+    def get_model_perms(self, request):
+        return {}
```
sarahfirstvoices, line 100: Can you please verify that this doesn't remove permission checking on the actual edit forms? E.g., staff should not have access to edit everything that superadmins can?

sarahfirstvoices (changes requested):

fv_be/app/admin/base_admin.py
sahil97: Yes, I could not find the official documentation for the has_model_perms method. From the [stackoverflow link](https://stackoverflow.com/questions/2431727/django-admin-hide-a-model), it seems it may have been used in the previous version of Django ?
I've updated it with another methods which I believe does the same thing and has been documentation for the current Django version.

fv_be/app/admin/base_admin.py
sahil97: Yes, I believe it does not remove permission checking on the edit forms.
I tested above by creating one more user (lets call it nonAdmin), gave it only permission to add a site and a dictionaryEntry and not other models. The nonAdmin then was only able to do the permitted stuff and was not able to CRUD on any other models which have also inherited from this class.

sarahfirstvoices (approved):

fv_be/app/admin/admin.py
```diff
@@ -70,6 +72,11 @@ class SiteAdmin(BaseAdmin):
 # todo: make suitable admin objects to register for each model
 # todo: make custom admin forms to prevent self selection in ManyToMany fields referring to self
 # ref: https://stackoverflow.com/questions/869856/
-admin.site.register(dictionary_models)
-admin.site.register(Category)
+admin.site.register(DictionaryEntry, DictionaryEntryHiddenBaseAdmin)
```
bronson-c, line 75: Would it be possible to use the @admin.register decorator in dictionary_admin.py rather than registering in this file?

NolanVH (approved):
All of these changes look great to me, thanks @sahil97!

One thing we might want to consider adding in a future PR is the addition of a DictionaryEntryAdmin class with a `list_filter` field to filter dictionary entries by site. It can be difficult to find and edit entries from the inline views only. This is probably something to discuss with others who will be using the admin site though.

fv_be/app/admin/admin.py
sahil97: Yes, great point. I added them here as of now because they all have the same baseAdmin class to register. I believe the decorator would come in handy once we start adding/making modifications to individual admin models which can then be registered in the dictionary_admin.py.

bronson-c (approved):

PR #19, opened 2023-03-23
FW-4121 Permissions with tests and sample API

NolanVH:
These changes add [Django Rules](https://github.com/dfunckt/django-rules) based permissions. The set of changes includes [base and view predicates, as well as some utilities](https://github.com/First-Peoples-Cultural-Council/fv-be/tree/d8b01e748cab6356277a7acb0c15f34f6b2302cf/fv_be/app/predicates). Additionally, there is a [large set of unit tests for the permissions](https://github.com/First-Peoples-Cultural-Council/fv-be/tree/d8b01e748cab6356277a7acb0c15f34f6b2302cf/fv_be/app/tests/test_predicates).

The complete set of tests can be executed using the command `pytest` from the root project directory. Currently, there are 290 total tests, which include tests for the Users app and tests for the permissions in this PR.

Along with the permissions, there is an example [API view for the site model](https://github.com/First-Peoples-Cultural-Council/fv-be/blob/d8b01e748cab6356277a7acb0c15f34f6b2302cf/fv_be/app/views/sites_views.py#L8-L22), which uses the permissions as defined in the site model meta classes [`rules_permissions`](https://github.com/First-Peoples-Cultural-Council/fv-be/blob/d8b01e748cab6356277a7acb0c15f34f6b2302cf/fv_be/app/models/sites.py#L86-L91) field.

Note that the permissions for staff and super admin roles are not fully implemented in these changes and will be completed in a separate ticket.


sahil97:
Great work. Doing some more testing locally but looks good to me.

sahil97 (approved):

bronson-c (approved):
Really great stuff, as Sahil mentioned there may be things that pop up during manual testing but this all LGTM 👍

PR #20, opened 2023-03-27
FW-4198 Add ViewSet mixin enabling list view permissions

NolanVH:
This PR adds a mixin which will enable rules-based permissions on model list views. If this mixin is used on a view, the permissions as defined in a model's meta class `rules_permissions` field will be used. The list view will return a list of all models of a type that the user has permission on and will leave out the rest.

Note: these changes depend on the changes in [FW-4121](https://github.com/First-Peoples-Cultural-Council/fv-be/pull/19) and should be merged after.

[FW-4121]: https://firstvoices.atlassian.net/browse/FW-4121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fv_be/app/views/base_views.py
```diff
@@ -0,0 +1,24 @@
+from rest_framework.response import Response
+from rules.contrib.rest_framework import AutoPermissionViewSetMixin
+
+
+class BaseViewSetMixin(AutoPermissionViewSetMixin):
```
sarahfirstvoices, line 5: This looks great! My only suggestion is to give this mixin a more descriptive name, like FVPermissionViewSetMixin or something

sarahfirstvoices (approved):

fv_be/app/views/base_views.py
bronson-c: I agree, the rest LGTM!

bronson-c (approved):

fv_be/app/views/base_views.py
NolanVH: Updated naming in https://github.com/First-Peoples-Cultural-Council/fv-be/pull/20/commits/4a6d78bdbb381cfb7cff5dcd4a7cd8df7227df02, thanks!

sahil97 (approved):

PR #21, opened 2023-03-27
FW-4181 Add script to automate local database cleaning, migrations, and superuser creation

NolanVH:
These changes add the following for ease of local development:

- [x]  Local database cleanup, migrations, and superuser creation script
- [x] Add information on how to use local database cleanup script to README
- [x] Add recommendations to README on how to setup linting for IDEs which matches pre-commit hook linting

### DIFF
--- a/src/reset-local-database.sh
+++ b/src/reset-local-database.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-echo "$PWD"
+# Get the path to this script
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
 # Using the input arguments, set the environment variables for a super-admin account username, password, and optional email.
 # Arguments are as follows:
@@ -89,12 +90,12 @@ case $yn in
       printf "Database creation failed: exit code $retval\n"
       exit $retval
     fi
-    
+
     # Note: this should be removed once we begin commiting migrations. See FW-4240
     # Delete any existing backend migrations files
     printf '\n\n'
     printf 'Removing existing backend migrations.\n'
-    find $PWD/src/firstvoices/backend/migrations/ -not -name '__init__.py' -delete
+    find $SCRIPT_DIR/firstvoices/backend/migrations/ -not -name '__init__.py' -delete
     retval=$?
     if [ $retval -ne 0 ]; then
       printf "Backend migration removal failed: exit code $retval\n"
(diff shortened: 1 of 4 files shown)


src/reset-local-database.sh
```diff
@@ -0,0 +1,160 @@
...
+    retval=$?
+    if [ $retval -ne 0 ]; then
+      printf "Migration execution failed: exit code $retval\n"
+      exit $retval
+    fi
+
+    # Create a superuser using the DJANGO_SUPERUSER_USERNAME, DJANGO_SUPERUSER_PASSWORD, and DJANGO_SUPERUSER_EMAIL environment variables.
+    printf '\n\n'
+    printf 'Creating a superuser account using the DJANGO_SUPERUSER_USERNAME, DJANGO_SUPERUSER_PASSWORD, and DJANGO_SUPERUSER_EMAIL environment variables.\n'
+    python $PWD/src/manage.py createsuperuser --noinput --id $DJANGO_SUPERUSER_USERNAME
```
sarahfirstvoices, line 146: Does this work with the current dev branch, where the user requires both an id and a username?

src/reset-local-database.sh
```diff
@@ -0,0 +1,162 @@
...
+    if [ $retval -ne 0 ]; then
+      printf "Database creation failed: exit code $retval\n"
+      exit $retval
+    fi
+    
+    # Note: this should be removed once we begin commiting migrations. See FW-4240
+    # Delete any existing backend migrations files
+    printf '\n\n'
+    printf 'Removing existing backend migrations.\n'
+    find $PWD/src/firstvoices/backend/migrations/ -not -name '__init__.py' -delete
```
bronson-c, line 97: I think you should remove the src from this path. As this script is in the `src` directory, the script tries to find the migrations at ... /fv-be/src/src/firstvoices/backend/migrations/

bronson-c: You should also remove the `src` from lines 108, 118, 128, 138 and 148.

bronson-c (changes requested):
The `src` should be removed from the file paths as it is included in `$PWD`. After this change the script works great!

src/reset-local-database.sh
NolanVH: That is a good point. The intention was to run the script from the root project directory. 

I have updated the script in https://github.com/First-Peoples-Cultural-Council/fv-be/commit/6c2ac2bba7b8a681ac090b6bfed9bdb78549b150 to get the path to the script and use that, instead of using the present working directory the script was executed from. It should now work regardless of where the user runs it from.

Thanks @bronson-c!

sarahfirstvoices (approved):

sahil97 (approved):

bronson-c (approved):

PR #22, opened 2023-03-28
fw-4191, merge dev to main

sarahfirstvoices:
My proposed order of operations here is to update the dev branch with these changes and then merge it into main with a merge commit.

Notes on the changes:

* I renamed the project and app folder to match the main branch, and added the extra 'firstvoices' parent folder because I'm guessing those paths are currently hooked up for ci/cd
* I kept the way dev had a local/production split for requirements and config, but updated the settings to match main. Super open to opinions and changes about this.
* I kept the pre-commit hooks from dev, so I kept the matching editorconfig (affects tabs v spaces and line lengths)
* I merged the users and auth stuff in a quick and dirty way that I think is roughly working, to keep the existing cognito integration and the existing admin setup working at the same time. There is a separate ticket logged to do this properly (fw-4165).
* I tried to pull out unused things from the generated template in dev-- if you spot more, comment and I'll take them out as well.


bronson-c:
Is there a purpose for keeping the `fv_be` folder or should it just be removed entirely?

firstvoices/firstvoices/users/api/serializers.py
NolanVH, line 1: It looks like [the user serializer](https://github.com/First-Peoples-Cultural-Council/fv-be/blob/c79c75cf4d475c90f2e41664d5512151b9b7b80d/firstvoices/firstvoices/users/api/serializers.py#L10) no longer matches the new fields after the user model changes, causing an error if the user view is accessed.

firstvoices/firstvoices/users/api/views.py
NolanVH, line 1: A small issue I noticed here is the users view returns an error because it is [asserting that the field is an integer in the get_queryset function](https://github.com/First-Peoples-Cultural-Council/fv-be/blob/c79c75cf4d475c90f2e41664d5512151b9b7b80d/firstvoices/firstvoices/users/api/views.py#L19). The field type was changed to a CharField.

NolanVH (changes requested):
These changes all look great, thanks @sarahfirstvoices!

Aside from the small user view and serializer issues everything seems to be working as intended.

One small addition that we might want to include in this PR is adding the migrations to the .gitignore while we are developing the new backend as we discussed.

NolanVH (approved):

sahil97 (approved):
LGTM

PR #23, opened 2023-03-29
FW-4197 Fixtures for partOfSpeech

sahil97:


README.md
```diff
@@ -32,6 +32,11 @@ License: Apache Software License 2.0
 1. Run the server and set it to automatically refresh on changes
    2. `npm install`
    3. `npm run dev`
+1. (Optional) To load data from fixtures, use the following command and replace the fixture_name with fixtures available.
+   1. `python manage.py loaddata fixture_name`
```
NolanVH, line 36: It might be worth mentioning that this command must be run from the `src` directory or the command must be modified to `python src/manage.py loaddata fixture_name` if in the root directory.

NolanVH (approved):
Tested and working locally. The fixtures load as expected. Thanks @sahil97!

src/firstvoices/backend/models/part_of_speech.py
```diff
@@ -6,24 +6,36 @@
...
+    """Manager to convert foreign key relationship to natural keys for fixtures to load correctly."""
+
+    def get_by_natural_key(self, title):
+        return self.get(title=title)
+
+
 class PartOfSpeech(BaseModel):
     """Model for Parts Of Speech."""
 
+    objects = ParentManager()
```
sarahfirstvoices, line 19: Doing this overrides the custom manager Nolan added to handle permissions, but since PartOfSpeech is not access-controlled I think it's alright

sarahfirstvoices (approved):

src/firstvoices/backend/models/base.py
```diff
@@ -55,3 +58,12 @@ class Meta:
 
     # from dc:lastContributor
     last_modified = models.DateTimeField(auto_now=True, db_index=True)
+
+
```
bronson-c, line 62: Does this function need to be called on the parts of speech? I noticed in the admin site I can still edit last created and last modified. If this will be called later to set those default values on parts of speech that's all good.

bronson-c (approved):
Tested locally, works well 👍 Just one question: In the admin site an admin must add the created_by and last_modified_by fields if they edit a part of speech. Is it possible for you to set default values in those fields?

sahil97:
> Tested locally, works well 👍 Just one question: In the admin site an admin must add the created_by and last_modified_by fields if they edit a part of speech. Is it possible for you to set default values in those fields?

Yes @bronson-c , So because of them being mandatory fields and being added automatically, the fixtures did not work directly. To get around the not-null violation, I had to add a pre_save_for_fixtures method in the base class, which checks if we are doing a 'raw' import, if yes, it will add the necessary dates (the date at which the data is being imported.).

src/firstvoices/backend/models/part_of_speech.py
sahil97: Now that I think of it, it can override the Permissions manager (I'll double check in the PartsOfSpeech API branch, once this gets merged).
My thinking of why it can override is that both managers are being assigned to the objects attribute. One way to avoid this would be to assign one of these managers to a specific attribute, for e.g. if we are assigning the permissions manager (coz of it being explicitly used in a queryset), to say 'permitted' attribute (or some more suitable name), then in the FVPermissionViewSetMixin, it can be called by `self.queryset.model.permitted.get_viewable_for_user(request.user) `.

@NolanVH your thoughts on this ?

NolanVH: In this case, it seems like we don't need both but if we do could the ParentManager not just inherit from PermissionsManager? That way we get the functionality from both and we also keep the `self.queryset.model.object.get_viewable_for_user(request.user)` intact.

PR #25, opened 2023-03-31
FW-4192 Add custom model manager for view permissions

NolanVH:
This changes adds a custom model manager `PermissionsManager` which is an extension of the built-in model manager. The new custom manager has a function `get_viewable_for_user` which takes in a user and will filter models based on that user's view permissions. Additional functionality can be added to this manager in the future, such as filtering based on other permissions.

Ex: `Site.objects.get_viewable_for_user(request.user)` will return a queryset containing all Site models that the requesting user has permission to view.

The `BaseModel` objects field has been set to use this manager, extending it's built in objects functionality. Any models which inherit from BaseModel can use this functionality.

### DIFF
--- a/src/firstvoices/backend/models/managers.py
+++ b/src/firstvoices/backend/models/managers.py
@@ -1,0 +1,20 @@
+from django.db import models
+
+
+class PermissionsManager(models.Manager):
+    """
+    This custom manager adds additional functionality to filter models based on user permissions. For example the
+    get_viewable_for_user function allows the filtering of model objects by the view permissions a user has.
+    """
+
+    def get_viewable_for_user(self, user):
+        """This function returns a queryset containing model objects that a user has view permissions on."""
+
+        # Get the list of model UUIDs that the user has view permissions for
+        models_uuid_list = [
+            model.id
+            for model in self.get_queryset()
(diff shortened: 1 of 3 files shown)


src/firstvoices/backend/models/managers.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+
+    def get_viewable_for_user(self, user):
+        """This function returns a queryset containing model objects that a user has view permissions on."""
+
+        # Get the list of model UUIDs that the user has view permissions for
+        models_uuid_list = [
+            model.id
+            for model in self.get_queryset()
+            if user.has_perm(f"backend.view_{model.__class__.__name__.lower()}", model)
```
sarahfirstvoices, line 17: I just realized that rules has a convenient function for getting the permission name-- this could be `user.has_perm(model.get_perm('view'))`. See: https://github.com/dfunckt/django-rules#permissions-in-models

NolanVH: Good catch, thanks. I have updated the check to use the function. Note that I still had to pass the model object so that the predicates have access to it: `user.has_perm(model.get_perm("view"), model)`

sarahfirstvoices (approved):

bronson-c (approved):

PR #26, opened 2023-04-03
FW-4172 G2P model integration + FW-4124 Reusable sort string field

bronson-c:
## This PR adds the following features:

- Adds support for G2P mappers to help with sorting and cleaning confusables.
    - G2P requires a default config, which is now stored as an AppJson model fixture.
    - G2P also requires the initialization of transducers, which is stored in the `AlphabetMapper` model in `characters.py`
- In order for the `AlphabetMapper` model to work correctly, the ManytoMany related dictionary entries field of the `Character` model was moved to `dictionary.py`.
- Adds an automatic custom sort string generated when a dictionary entry is created/updated.

## To set up custom order and confusable cleaning for dictionary entries:
- Load the `default_g2p_config.json` fixture
- Create `Character` models that correspond with the characters you will use in a site's alphabet.
    - Base characters are require, ignorables and variants are optional. 
- Create an `AlphabetMapper` class with an appropriate input to canonical mapping that defines confusables.  
    - For example: ```[{"in": "á", "out": "a"}, {"in": "à", "out": "a"}, {"in": "ᐱ", "out": "A"}, {"in": "Á", "out": "A"}, {"in": "À", "out": "A"}, {"in": "c̣", "out": "c"}, {"in": "č", "out": "c"}, {"in": "C̣", "out": "C"}, {"in": "Č", "out": "C"}, {"in": "ȼh", "out": "ch"}, {"in": "Ȼh", "out": "Ch"}]```


src/firstvoices/backend/models/characters.py
```diff
@@ -143,21 +146,43 @@ def __str__(self):
...
+
+    class Meta:
+        verbose_name = _("confusable mapper")
+        verbose_name_plural = _("confusable mappers")
+
+    # from fv-alphabet:confusable_characters
+    # JSON representation of a g2p mapping from confusable characters to canonical characters
+    input_to_canonical_map = models.JSONField()
+    # TODO: Possibly write a custom field that takes YAML here
+    g2p_config_yaml = models.TextField(blank=True)
```
sarahfirstvoices, line 162: You could store this as JSON internally and convert it to yaml when needed using the standard json and yaml modules: https://www.pythonforbeginners.com/basics/convert-json-to-yaml-in-python

src/firstvoices/backend/models/characters.py
```diff
@@ -143,21 +146,43 @@ def __str__(self):
...
+    def generate_g2p_config(self):
+        """Generates the site-specific yaml config file using a yaml template."""
+
+        site_name = f"FV {self.site.title}"
+        site_code = f"fv-{self.site.slug}"
+
+        # TODO: default_config yaml storage location is temporary
+        default_yaml_path = os.path.join(
+            os.path.dirname(os.path.abspath(__file__)),
+            "../utils/characterfiles/default_config.yaml",
```
sarahfirstvoices, line 180: One place to potentially keep this is the AppJson model, where we also have the default site menu. (Again, converting to yaml as needed.)

src/firstvoices/backend/models/characters.py
```diff
@@ -1,9 +1,11 @@
+import g2p
```
NolanVH, line 1: The g2p package should probably be added to the base requirements file.

src/firstvoices/backend/models/dictionary.py
```diff
@@ -140,13 +142,44 @@ class TypeOfDictionaryEntry(models.TextChoices):
...
     class Meta:
         verbose_name = _("Dictionary Entry")
         verbose_name_plural = _("Dictionary Entries")
 
     def __str__(self):
         return self.title
 
+    def save(self, *args, **kwargs):
+        self.clean_confusables()
+        self.set_custom_order()
```
NolanVH, line 162: It might be an idea to do a check here and not call these functions if confusables don't exist for a site.

NolanVH (changes requested):
These changes are awesome! Thanks, @bronson-c and @caforbes!

I found a few minor issues described in the comments below otherwise, this is looking great.

src/firstvoices/backend/models/characters.py
bronson-c: I can add g2p to the base requirements file but currently there is a Werkzeug dependency conflict between our repo and the most recent versions of g2p. Currently installing g2p via the requirements file with an unspecified version installs g2p version 0.5.20220318

src/firstvoices/backend/models/dictionary.py
```diff
@@ -140,13 +142,44 @@ class TypeOfDictionaryEntry(models.TextChoices):
...
+    def clean_confusables(self):
+        mapper = AlphabetMapper.objects.filter(site_id=self.site_id).first()
+        confusables_transducer = mapper.preprocess_transducer if mapper else {}
+
+        # strip whitespace and normalize
+        self.title = self.title.strip()
+        self.title = nfc(self.title)
+
+        # clean confusables
+        cleaned_title = confusables_transducer(self.title).output_string
```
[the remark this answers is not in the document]
caforbes: I would like to update this so that a mapper for the site is created on first access if possible. The mapper needs to exist in order to run the sort process on any entry.

src/firstvoices/backend/models/characters.py
caforbes: Message from Aidan/g2p is that they are sticking with werkzeug 2.0 in order to support python 3.6. So to resolve this, we either need to downgrade or look into an alternative dev server option like gunicorn.

requirements/local.txt
```diff
@@ -1,6 +1,5 @@
 -r base.txt
 
-Werkzeug[watchdog]==2.2.3 # https://github.com/pallets/werkzeug
```
bronson-c, line 3: Removed the requirement so the version required for G2P is installed automatically

sarahfirstvoices:
> To set up custom order and...

This is great info that should live in a permanent place. A readme section for now?

src/firstvoices/backend/models/characters.py
```diff
@@ -143,21 +103,123 @@ def __str__(self):
...
+        """
+        if self.input_to_canonical_map != {}:
+            preprocess_settings = self.default_g2p_config["preprocess_config"]
+
+            return g2p.Transducer(
+                g2p.Mapping(**preprocess_settings, mapping=self.input_to_canonical_map)
+            )
+        else:
+            self.logger.warning("No confusable map found for site %s", self.site)
+            return None
```
sarahfirstvoices, line 174: Just to confirm: the sorting can still function with no preprocess transducer?

src/firstvoices/backend/models/dictionary.py
```diff
@@ -140,13 +142,39 @@ class TypeOfDictionaryEntry(models.TextChoices):
...
 
+    def save(self, *args, **kwargs):
+        mapper, created = AlphabetMapper.objects.get_or_create(site_id=self.site_id)
+
+        self.clean_title(mapper)
+        self.set_custom_order(mapper)
+        super().save(*args, **kwargs)
+
+    def clean_title(self, mapper):
+        # strip whitespace and normalize
```
sarahfirstvoices, line 168: Would it make sense to move this function to the AlphabetMapper and make it return the cleaned string? It could be called something generic like normalize (or maybe.... ultranormalize). It needs the mapper anyway, and it seems like the kind of cleaning operation we may want to reuse for things like search query inputs or song titles.

src/firstvoices/backend/models/characters.py
```diff
@@ -143,21 +103,123 @@ def __str__(self):
...
-#
-#     # See fw-4172: Once file storage is configured, enable these FileFields / change to proper paths
-#     # alphabet_mapper is a placeholder for now.
-#     # input_to_canonical = models.FileField(upload_to="alphabet_mapper")
-#     # canonical_to_base = models.FileField(upload_to="alphabet_mapper")
-#     # g2p_config = models.FileField(upload_to="alphabet_mapper")
-#
-#     def __str__(self):
-#         return self.id
+class AlphabetMapper(BaseSiteContentModel):
```
sarahfirstvoices, line 106: This has enough different properties and methods that I wonder if it would make sense to call it just Alphabet

src/firstvoices/backend/models/characters.py
```diff
@@ -143,21 +103,123 @@ def __str__(self):
...
+    """
+    Hosts text processors and sorters for a site's custom alphabet.
+    Must be re-created upon changing any character models in an alphabet.
```
sarahfirstvoices, line 109: I'm sure we've talked about this before so I'm sorry for asking, but why does this need to be re-created after changing any character models? It looks like the only thing that is stored is the input_to_canonical_map. If I added a character or changed the characters' sort orders, wouldn't this just keep working with the existing confusable list?

src/firstvoices/backend/utils/CharacterSortOrderHelpers.py
```diff
@@ -0,0 +1,127 @@
+import re
```
sarahfirstvoices, line 1: Let's stick to our file naming conventions and call this something like character_utils.py

src/firstvoices/backend/utils/CharacterSortOrderHelpers.py
```diff
@@ -0,0 +1,127 @@
...
+import unicodedata
+
+
+# From MTD processors
```
sarahfirstvoices, line 5: For the benefit of our future selves, can you please include a more detailed comment about where this came from? Preferably with a permalink to the relevant line in a git repo

src/firstvoices/backend/utils/CharacterSortOrderHelpers.py
```diff
@@ -0,0 +1,127 @@
...
+
+    Adds space as first alphabet item. Defines a map from integer values
+    to custom sort characters as a string. Adds a flag before out-of-vocab characters.
+
+    Examples:
+        >>> sorter = CustomSorter(['a', 'b', 'c'])
+        >>> sorter.word_as_values('ab abcd')
+        [1, 2, 0, 1, 2, 3, 10100]
+        >>> sorter.word_as_sort_string('ab abcd')
+        '!# !#$⚑d'
```
sarahfirstvoices, line 82: Very helpful docstring. Thanks for this.

sarahfirstvoices:
Looking forward to some tests for this. Will that be in a separate review?

src/firstvoices/backend/models/characters.py
bronson-c: Yes, when there is no transducer the clean confusables function of the alphabet mapper simply does not process the text and returns it.

bronson-c:
> Looking forward to some tests for this. Will that be in a separate review?

@caforbes is working on the tests, I believe that they will be in this PR

src/firstvoices/backend/models/dictionary.py
caforbes: The `clean_confusables` part of this function is only necessary for sorting, and is on the AlphabetMapper already. We don't currently apply this to any non-sorted fields, so I would hesitate to extend it. But the NFC and stripping operations could definitely get pulled out to a more generic place so that we can later apply them to all user input or text fields.

src/firstvoices/backend/models/characters.py
bronson-c: Yes you are correct, I will update the comment. The only thing that needs manual adjustment is after an alphabet is edited the confusables map on the alphabet mapper should be updated to include any confusable characters.

NolanVH (commented):
This is looking great. Good work on getting the dependency issue resolved and the pytest logging issue fixed!

I ran everything locally and it all seems to be working except for a minor issue with the existing tests. It looks like the tests query for the default_g2p_config but the fixture is not ever loaded in the test database causing them to fail. I believe the fixture can be autoloaded into the test database with the following addition to the [conftest.py](https://github.com/First-Peoples-Cultural-Council/fv-be/blob/27dede2beb171e2b12165e488166a8ff9dbcc4e8/src/firstvoices/backend/tests/conftest.py) file:
```
@pytest.fixture(scope='session')
def django_db_setup(django_db_setup, django_db_blocker):
    with django_db_blocker.unblock():
        call_command('loaddata', 'default_g2p_config.json')
```

Additionally, it might be useful to add a `loaddata` command to the reset local database script to autoload the g2p config fixture.

src/firstvoices/backend/tests/test_models/test_character_model.py
```diff
@@ -0,0 +1,177 @@
...
+class TestCharacterModel:
+    @pytest.mark.django_db
+    def test_characters_same_name(self):
+        """Character can't be created with the same name as another"""
+        char = CharacterFactory.create()
+        with pytest.raises(IntegrityError):
+            CharacterFactory.create(site=char.site, title=char.title)
+
+    @pytest.mark.django_db
+    def test_character_variant_same_name(self):
```
sarahfirstvoices, line 138: This just a note for the future but if you want to easily test all combos of these factories, the test parameter feature lets you multiple two lists together

sarahfirstvoices (approved):

bronson-c:
> Looks good. Some optional suggestions:
> 
>     * Separate files for factories for different models
> 
>     * Separate files for testing Character and Alphabet models
> 
>     * Absolute imports instead of relative imports

As per the pep 8 discussion the imports will stay as-is. I've separated the character and alphabet tests. Moving factories should be its own small ticket since there are many factories that fit into many different categories in that file.

PR #29, opened 2023-04-04
Fw 4228 read api for parts of speech list and detail

sahil97:


src/firstvoices/backend/views/dictionary_views.py
```diff
@@ -0,0 +1,24 @@
...
+from firstvoices.backend.views.base_views import FVPermissionViewSetMixin
+
+
+@extend_schema_view(
+    list=extend_schema(
+        description="A list of different parts-of-speech available that different words/phrases can be assigned to.",
+        responses={200: PartsOfSpeechSerializer},
+    ),
+    retrieve=extend_schema(
+        description="Details about a specific parts-of-speech.",
```
sarahfirstvoices, line 17: Typo: singular part of speech here, and I don't think it needs to be hyphenated.

src/firstvoices/backend/tests/test_dictionary_models/test_parts_of_speech.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+# Fixtures
+@pytest.fixture(scope="session")
+def parts_of_speech_db_setup(django_db_setup, django_db_blocker):
+    with django_db_blocker.unblock():
+        call_command("loaddata", "partsOfSpeech_initial.json")
+
+
+class TestPartsOfSpeechModel:
```
sarahfirstvoices, line 22: Can you please separate the model tests from the API tests?

src/firstvoices/backend/tests/test_dictionary_models/test_parts_of_speech.py
```diff
@@ -0,0 +1,70 @@
...
+        ), "Parent's title does not match"
+
+    def test_fixture_with_model(self, db, parts_of_speech_db_setup):
+        # Testing that the fixture works, and we have at least 1 object in the database
+        pos_count = get_list_or_404(PartOfSpeech)
+        assert len(pos_count) != 0
+
+
+class TestPartsOfSpeechView:
+    """Tests for parts-of-speech views."""
```
sarahfirstvoices, line 56: Are there more tests coming for the API?

src/firstvoices/backend/models/part_of_speech.py
```diff
@@ -29,6 +33,12 @@ class PartOfSpeech(BaseModel):
     class Meta:
         verbose_name = _("Part Of Speech")
         verbose_name_plural = _("Parts Of Speech")
+        rules_permissions = {
+            "view": rules.always_allow,
+            "add": predicates.is_superadmin,
+            "change": predicates.is_superadmin,
+            "delete": predicates.is_superadmin,
+        }
```
sarahfirstvoices, line 41: +1 for these permission settings. Looks good.

src/firstvoices/backend/views/dictionary_views.py
```diff
@@ -0,0 +1,24 @@
...
+        responses={200: PartsOfSpeechSerializer},
+    ),
+)
+class PartsOfSpeechViewSet(FVPermissionViewSetMixin, ModelViewSet):
+    http_method_names = ["get"]
+    serializer_class = PartsOfSpeechSerializer
+    queryset = PartOfSpeech.objects.all()
```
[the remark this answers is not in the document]
gmcauliffe: Currently we're not doing anything with the "parent" in the FE, so a flat list would be sufficient for our current use cases. I can see a nested structure being helpful in the future though.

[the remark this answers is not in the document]
sahil97: No specific reason, but I was going with a similar structure to what we had before. I agree that nesting would be more suitable. I'll add that.

src/firstvoices/backend/tests/test_dictionary_models/test_parts_of_speech.py
sahil97: I don't have any specific ideas as of now, but would like to add more tests to make it more well tested.

src/firstvoices/backend/models/part_of_speech.py
NolanVH: Is there any situation where a staff user might need to add to or edit parts of speech?

src/firstvoices/backend/tests/test_dictionary_models/test_parts_of_speech.py
NolanVH: It might be worth testing that each user type, (superadmin, staff, language admin, editor, assistant, member, anonymous), can see the parts of speech, duplicate titles aren't possible, and testing that grandparent creation is not possible.

NolanVH (commented):
These changes look great, thanks @sahil97!

A few minor comments listed below. Additionally, it might be a good time to update the admin panel so that the created by and modified by fields are automatically added for parts of speech.

This can be done very easily by creating a PartsOfSpeechAdmin class which inherits the BaseAdmin class. Something like the following:
```
class PartOfSpeechAdmin(BaseAdmin):
    list_display = (
        "title",
    ) + BaseAdmin.list_display
```

src/firstvoices/backend/models/part_of_speech.py
NolanVH: Is there any situation where a staff user might need to add to or edit parts of speech?

src/firstvoices/backend/tests/test_dictionary_models/test_parts_of_speech.py
sahil97: Thanks @NolanVH for the ideas, these can be great additions. Since there is not Create/Update/Delete as of now for this view, so I didn't add some of the tests you have mentioned as of now. I will add the grandparent creation check and a test to check if diff. user roles can see the parts of speech though which seems like good additions.

sarahfirstvoices: It would also be good to test that the data format is as expected, since there aren't separate tests for the serializer. At a minimum, the number of items in the returned list and the property names for one of them.

bronson-c (commented):
Tested locally and everything works well 👍 will approve once the discussed changes to tests and nesting are made

src/firstvoices/backend/tests/test_dictionary/test_parts_of_speech_view.py
```diff
@@ -0,0 +1,114 @@
...
+    API_LIST_VIEW = "api:parts-of-speech-list"
+    API_DETAIL_VIEW = "api:parts-of-speech-detail"
+    APP_NAME = "backend"
+
+    @pytest.fixture(scope="session")
+    def parts_of_speech_db_setup(self, django_db_setup, django_db_blocker):
+        with django_db_blocker.unblock():
+            call_command("loaddata", self.FIXTURE_FILE)
+
+    def test_base_case_list_view(self, db, parts_of_speech_db_setup):
```
NolanVH, line 25: We could parameterize this test and others to check multiple user types instead of just the staff role. Ex: https://github.com/First-Peoples-Cultural-Council/fv-be/blob/d575ba2be3ed3f21a751f703b42e7c3b60ee7522/firstvoices/backend/tests/test_predicates/test_predicate_base.py#L121-L128

src/firstvoices/backend/admin/dictionary_admin.py
```diff
@@ -64,3 +64,7 @@ class DictionaryEntryHiddenBaseAdmin(HiddenBaseAdmin):
         AlternateSpellingInline,
         PronunciationInline,
     ]
+
+
+class PartsOfSpeechAdmin(BaseAdmin):
+    list_display = ("title",) + BaseAdmin.list_display
```
NolanVH, line 70: The "parent" field could be added here.

NolanVH (commented):
These changes are looking great, thanks @sahil97! A few minor suggestions and then I think this is ready.

bronson-c (approved):
LGTM! 👍  (after you make the changes Nolan suggested 🙂 )

src/firstvoices/backend/serializers/dictionary_serializers.py
```diff
@@ -0,0 +1,17 @@
...
+        model = PartOfSpeech
+        fields = ["id", "title"]
+
+
+class PartsOfSpeechSerializer(serializers.ModelSerializer):
+    children = PartsOfSpeechChildrenSerializer(many=True)
+
+    class Meta:
+        model = PartOfSpeech
+        fields = ["id", "title", "children"]
```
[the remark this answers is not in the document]
sarahfirstvoices: Optional: the more detailed serializer could subclass the less detailed one to avoid repetition. (E.g., see the SiteSummary and SiteDetail serializers)

src/firstvoices/backend/views/base_views.py
```diff
@@ -14,9 +14,13 @@ class FVPermissionViewSetMixin(AutoPermissionViewSetMixin):
     model permissions are registered using ``rules.contrib.models.RulesModelMixin``.
     """
 
+    # Method to override the list queryset to implement different querysets for list and retrieve respectively
+    def get_list_view_queryset(self):
```
sarahfirstvoices, line 18: Thanks for adding this. It's going to be so handy. We already have 2 cases where the list and detail views have different querysets!

src/firstvoices/backend/admin/dictionary_admin.py
sahil97: Hmm, I'm not sure why but I can still see and am able to edit the parent field without mentioning it here. I've also tried hard reloading to eliminate the cache possibility on browser side, is there any other reason why "parent" field is still visible without mentioning here ?
I've added it explicitly now though, but just wondering.

src/firstvoices/backend/views/dictionary_views.py
```diff
@@ -0,0 +1,28 @@
...
+
+    @staticmethod
+    def get_list_queryset():
+        return PartOfSpeech.objects.exclude(parent__isnull=False)
```
sarahfirstvoices, line 28: Does this result in a bunch of separate queries to get the child parts of speech? Do we need any prefetching / select related options here?

src/firstvoices/backend/tests/test_dictionary/test_parts_of_speech_view.py
```diff
@@ -0,0 +1,114 @@
...
+        client = APIClient()
+        member_user = get_app_admin(AppRole.STAFF)
+        client.force_authenticate(user=member_user)
+        response = client.get(reverse(self.API_LIST_VIEW, current_app=self.APP_NAME))
+        content = json.loads(response.content)["results"]
+
+        assert isinstance(content, list)
+        assert len(content) > 0
+
+        print(content)
```
NolanVH, line 103: This print statement should probably be removed.

src/firstvoices/backend/admin/dictionary_admin.py
NolanVH: The "list_display" attribute only affects the list view in the admin panel (eg: "/admin/backend/partofspeech/"). If the "fields" attribute is not given here then for the create and edit pages all fields will be displayed.

src/firstvoices/backend/tests/test_dictionary/test_parts_of_speech_view.py
sahil97: Yes, my bad. I suggest we should also add a check in the pre-commit hooks to not let this go through.

sarahfirstvoices (approved):

PR #32, opened 2023-04-13
Fw 4252 complete the list of parts of speech in fixtures

sahil97:


src/firstvoices/backend/fixtures/partsOfSpeech_initial.json
```diff
@@ -261,5 +261,69 @@
...
+      "title": "Reflexive Verb",
+      "parent": [
+        "Verb"
+      ]
+    }
+  },
+  {
+    "model": "backend.partofspeech",
+    "fields": {
+      "title": "Entity",
```
sarahfirstvoices, line 284: From the wiki page it looks like this should be "Entity (Noun-like word)

src/firstvoices/backend/fixtures/partsOfSpeech_initial.json
```diff
@@ -261,5 +261,69 @@
...
+    "model": "backend.partofspeech",
+    "fields": {
+      "title": "Word Piece/Affix",
+      "parent": null
+    }
+  },
+  {
+    "model": "backend.partofspeech",
+    "fields": {
+      "title": "Event/Activity",
```
sarahfirstvoices, line 298: From the wiki page it looks like this should be "Event/Activity (Verb-like word)"

sarahfirstvoices: Also on the wiki page this lists itself as a parent... I wonder if the parent is meant to be "Verb"?

sarahfirstvoices (changes requested):
Comparing to the wiki page, a couple of the existing rows are also different. Could you please double check the title of "modifier verb" and the parent of "demonstrative"?

src/firstvoices/backend/fixtures/partsOfSpeech_initial.json
sahil97: Oh yes, this does not have a parent. Have confirmed from @caforbes and have updated the wiki page as well.

NolanVH (approved):

sarahfirstvoices (approved):

caforbes (approved):
Looking good!

PR #33, opened 2023-04-14
FW-4229 Default categories on new site creation

NolanVH:
### Description of Changes
These changes add functionality to the site model save function to auto-add a set of default categories to any newly created sites. The category model "description" field is now also optional and the default categories have empty descriptions.

Additionally, these changes needed the use of the "Category" model in the "Site" model. This was causing a circular import issue so I moved the "BaseSiteContentModel" and "BaseControlledSiteContentModel" files to the base.py file and updated the "site" foreign key on "BaseSiteContentModel" to use `to="backend.Site"` instead of importing "Site" and using it directly:  https://github.com/First-Peoples-Cultural-Council/fv-be/blob/fb8d95f3c8e81c4d6c13eb229bc11e40f6359732/src/firstvoices/backend/models/base.py#L78-L80

### Checklist

### Additional Notes
N/A

### DIFF
--- a/src/firstvoices/backend/models/utils/default_categories.json
+++ b/src/firstvoices/backend/models/utils/default_categories.json
@@ -1,0 +1,268 @@
+[
+    {
+        "model": "backend.category",
+        "fields": {
+            "title": "Animals",
+            "parent": ""
+        }
+    },
+    {
+        "model": "backend.category",
+        "fields": {
+            "title": "Birds",
+            "parent": "Animals"
+        }
+    },
+    {
+        "model": "backend.category",
+        "fields": {
+            "title": "Body",
+            "parent": ""
+        }
+    },
+    {
+        "model": "backend.category",
+        "fields": {
+            "title": "Body Parts",
+            "parent": "Body"
+        }
+    },
+    {
+        "model": "backend.category",
+        "fields": {
+            "title": "Buildings",
+            "parent": "Human Things / Activities"
+        }
+    },
+    {
+        "model": "backend.category",
+        "fields": {
+            "title": "Colours",
+            "parent": ""
+        }
+    },
+    {
+        "model": "backend.category",
+        "fields": {
+            "title": "Cultural Objects & Practices",
+            "parent": "Human Things / Activities"
+        }
+    },
+    {
+        "model": "backend.category",
+        "fields": {
+            "title": "Essential Phrases",
+            "parent": ""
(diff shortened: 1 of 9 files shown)


src/firstvoices/backend/models/utils/default_categories.json
```diff
@@ -0,0 +1,268 @@
...
+        "fields": {
+            "title": "Food",
+            "parent": ""
+        }
+    },
+    {
+        "model": "backend.category",
+        "fields": {
+            "title": "Fungi & Lichens",
+            "parent": "Plants"
```
sarahfirstvoices, line 97: Ha ha, this is my classification pet peeve, but also I'm sure it's fine.

src/firstvoices/backend/models/utils/utils.py
```diff
@@ -0,0 +1,60 @@
...
+
+from firstvoices.backend.models.category import Category
+
+
+def load_data(json_file):
+    """Helper function to load data from a json file."""
+
+    with open(path.dirname(__file__) + "/" + json_file, encoding="utf-8") as data_file:
+        json_data = json.loads(data_file.read())
+        return json_data
```
sarahfirstvoices, line 12: On the one hand, this totally works. On the other hand, we are already storing some default json values in the AppJson table (default site menu, and the default g2p config). Maybe it would make sense to add this there as well, so it's all together?

sahil97: Looks good to me either way. I believe this separation of logic makes this more flexible, in case we need to change this later or call the categories generation from a mock site generation function. Maybe I can add a ticket to merge this into AppJson later but this works from the MVP perspective.

caforbes (commented):
Looks good to me, but wondering if there should be any tests for the onsave behaviour or util functions?

src/firstvoices/backend/models/utils/utils.py
sarahfirstvoices: Hmm, I think this way is actually less flexible-- if we ever want to change the default categories we will have to deploy a new file, instead of just updating it in the database

sahil97: Sounds good. I've added a ticket for this in the Backend delighters epic with a brief details about this for us to come back to this.

sarahfirstvoices (approved):

PR #38, opened 2023-04-22
fw-4191, add remaining work from dev

sarahfirstvoices:
- added remaining models, view, serializers, etc
- add settings_test.py to get test users working in automated tests
- altered the User model to work with the permissions-- I could still log in via OAuth but Rob please check the changes
- enabled linting and precommit hooks, and formatted existing files
- added a migration


firstvoices/backend/urls.py
```diff
@@ -1,24 +1,20 @@
...
 
 urlpatterns = [
-	path(r'health', HealthCheckView.as_view()),
-	# path("api/schema/", SpectacularAPIView.as_view(), name="api-schema"),
-	# path(
-	# 	"api/docs/",
-	# 	SpectacularSwaggerView.as_view(url_name="api-schema"),
-	# 	name="api-docs",
-	# ),
+    path(r"health", HealthCheckView.as_view()),
```
sarahfirstvoices, line 17: I realize this changes the url of the health check. My proposal is to deal with that as its own issue AFTER the merge

plasticviking (approved):

PR #45, opened 2023-04-25
Fw-4200, dictionary API

sarahfirstvoices:
### Description of Changes
Implementation of a basic dictionary API (list and detail view) with tests and docs. There are some extra changes as well:

* Base classes for API tests
* Support for nested URL config (e.g., site/site-name/dictionary/entry-id)
* Reusable serializer fields for working with nested URLs
* Some updates to dictionary models, to use the standard base classes
* Adjusted the current user API to work with the API doc generation

Not included:

* related entries data in the responses, because I want to deal with the related object permissions in its own PR
* changing the DictionaryEntry's foreign key on categories to a many-to-many in order to multiple categories (will do in separate PR)

### Checklist
- [x] README / documentation has been updated if needed
- [x] PR title / commit messages contain Jira ticket number if applicable
- [x] Tests have been updated / created if applicable
- [x] Admin Panel has been updated if models have been added or modified
- [x] Migrations have been updated and committed if applicable

### Reviewers Should Do The Following:
- [x] Review the code changes here
- [x] Pull the branch and test locally -- including the docs and the response data format, please

PR #46, opened 2023-04-26
FW-4241 FW-4240 Add add default fixture migration and update reset script

NolanVH:
### Description of Changes
These changes add two migrations which do the following:

In the first migration:
1.  Set null field to True for the `created` and `last_modified` fields on the `AppJson` and `PartsOfSpeech` models (since the fixtures don't have these values).
2. Load the entries for the default fixtures.
3. Update the `created` and `last_modified` fields on the loaded entries to the current time.

In the second migration:
1. Set the null field back to False for the `created` and `last_modified` fields on the `AppJson` and `PartsOfSpeech` models (the revert of the AlterField operation must be done in a separate migration).

Additionally, the `reset-local-database.sh` script has been updated with the following changes:
- migrations are no longer deleted now that they are being committed to the repo
- the `dropdb` commands have the `--if-exists` flag added to no longer error if the databases can not be found (will log a message)
- The fixtures are no longer loaded in the script (since they are now loaded in the migration)

### Checklist

- [x] Pull the branch and test locally - specifically test the reset-local-database.sh script if possible

### Additional Notes
Will rebase this with the dictionary API migrations changes after https://github.com/First-Peoples-Cultural-Council/fv-be/pull/45 has been merged.


sahil97 (approved):
Tested Locally. Works as expected.

PR #50, opened 2023-04-27
FW-4257 Add my-sites read API and tests

NolanVH:
### Description of Changes
This PR adds the `my-sites` API which returns a list of language sites that the given user has membership to and permission to view.

The view only contains the list operation at `/api/1.0/my-sites/` and the retrieve operation for single sites is disabled, as `/api/1.0/sites/{site-slug}/` can be used for details. The list view of the my-sites endpoint contains a URL field for each returned site which links to this detail endpoint.

The list view returns the `sites` data grouped by language. It uses the `SiteSummarySerializer` to serialize and format the data. When reviewing, please let me know if any changes to the returned data are needed or might be useful in the future.

### Checklist

### Additional Notes
N/A


firstvoices/backend/tests/test_apis/test_sites_api.py
```diff
@@ -186,3 +186,142 @@ def test_detail_404(self):
...
+class TestMySites(BaseApiTest):
+    """
+    End to end tests that check the my-sites endpoint for expected behavior.
+    """
+
+    API_LIST_VIEW = "api:my-sites-list"
+    APP_NAME = "backend"
+
+    @pytest.mark.django_db
+    def test_list_empty(self):
```
bronson-c, line 200: Based on the changes to the base API test class I think that this test can be removed

bronson-c: Nevermind, I see this is a different case

bronson-c (approved):

firstvoices/backend/tests/factories.py
```diff
@@ -178,6 +178,12 @@ def get_site_with_member(site_visibility, user_role):
     return site, user
 
 
+def get_site_with_user_as_member(site_visibility, user, user_role):
```
sarahfirstvoices, line 181: Would the existing get_site_with_member work for this same case? It returns a site and user

firstvoices/backend/tests/test_apis/test_sites_api.py
```diff
@@ -186,3 +186,142 @@ def test_detail_404(self):
         response = self.client.get(f"{self.get_detail_endpoint('fake-site')}")
 
         assert response.status_code == 404
+
+
+class TestMySites(BaseApiTest):
```
sarahfirstvoices, line 191: These are great but please put them in their own file. my-sites is its own api :)

firstvoices/backend/views/sites_views.py
```diff
@@ -96,3 +96,76 @@ def list(self, request, *args, **kwargs):
...
         return Response(data)
+
+
+@extend_schema_view(
+    list=extend_schema(
+        description="A list of language sites a given user is a member of, "
+        "grouped by language. If there are no accessible sites or "
+        "the user does not have membership to any site then the "
+        "list will be empty. Sites with no specified language "
+        "will be grouped under 'Other'.",
```
sarahfirstvoices, line 107: My only question about this is whether we want the sites grouped by language or just in a plain list. @gmcauliffe do you have thoughts?

firstvoices/backend/views/sites_views.py
```diff
@@ -96,3 +96,76 @@ def list(self, request, *args, **kwargs):
...
+        # get the site ids from the memberships
+        for membership in memberships:
+            sites_id_list.append(membership.site.id)
+
+        # get the site objects filtered by the id list and order by language
+        queryset = (
+            Site.objects.filter(id__in=sites_id_list)
+            .select_related("language")
+            .order_by(Upper("language__title"), Upper("title"))
+        )
```
sarahfirstvoices, line 144: I think this could all be done in one queryset. Something like Membership.objects.filter(user=x).select_related('site', 'site__language__title').order_by('site__language__title')

firstvoices/backend/views/sites_views.py
```diff
@@ -96,3 +96,76 @@ def list(self, request, *args, **kwargs):
...
+        # group by language
+        rows = groupby(site_jsons, itemgetter("language"))
+        data = [
+            {
+                "language": language if language is not None else "Other",
+                "sites": list(items),
+            }
+            for language, items in rows
+        ]
+
```
sarahfirstvoices, line 170: Can this be extracted to a reusable function that is shared with the SiteViewSet?

firstvoices/backend/urls.py
```diff
@@ -3,13 +3,14 @@
...
 from backend.views.parts_of_speech_views import PartsOfSpeechViewSet
-from backend.views.sites_views import SiteViewSet
+from backend.views.sites_views import MySitesViewSet, SiteViewSet
 from backend.views.user import UserViewSet
 
 # app-level APIs
 ROUTER = DefaultRouter(trailing_slash=True)
 ROUTER.register(r"user", UserViewSet, basename=r"user")
 ROUTER.register(r"parts-of-speech", PartsOfSpeechViewSet, basename="parts-of-speech")
+ROUTER.register(r"my-sites", MySitesViewSet, basename="my-sites")
```
sarahfirstvoices, line 13: Are you doing something to disable the detail view? I think we could register a single view here rather than the whole viewset

firstvoices/backend/views/sites_views.py
```diff
@@ -96,3 +96,76 @@ def list(self, request, *args, **kwargs):
...
+                    "language": serializers.CharField(),
+                    "sites": SiteSummarySerializer(many=True),
+                },
+            ),
+        },
+    ),
+)
+class MySitesViewSet(
+    AutoPermissionViewSetMixin, mixins.ListModelMixin, viewsets.GenericViewSet
+):
```
sarahfirstvoices, line 121: Oh, I get it. Only adding the list mixin. That works! Ignore my comment about only registering the list view in the urls

sarahfirstvoices (approved):
I'm clicking approve because I only suggested minor changes. I think they're worth doing but no need to re-review after :)

firstvoices/backend/tests/factories.py
NolanVH: I think in this case it wouldn't work as some of the tests need a single user with membership to multiple sites. The `get_site_with_member` function creates a separate user with each call.

I have added an optional user parameter to the `get_site_with_member` function to add this functionality, and I have removed `get_site_with_user_as_member`.

firstvoices/backend/views/sites_views.py
NolanVH: Good suggestion. Thanks!

firstvoices/backend/views/sites_views.py
sarahfirstvoices: I chatted with @gmcauliffe and it would be better if this were all in one list. No need to group by language

sarahfirstvoices (changes requested):

firstvoices/backend/views/sites_views.py
NolanVH: I have removed the grouping by language for this API. Thanks for the feedback!

firstvoices/backend/views/sites_views.py
```diff
@@ -72,27 +93,53 @@ def list(self, request, *args, **kwargs):
...
+    http_method_names = ["get"]
+    pagination_class = None
+    serializer_class = SiteSummarySerializer
+
+    def get_queryset(self):
+        # get the site objects filtered by the membership set for the user
+        # note that the titles are converted to uppercase and then sorted which will put custom characters at the end
+        queryset = Site.objects.filter(membership_set__user=self.request.user).order_by(
+            Upper("title")
+        )
```
sarahfirstvoices, line 141: Probably still want to select_related(language) since it is used in the serializer

sarahfirstvoices (approved):

firstvoices/backend/views/sites_views.py
NolanVH: Good point. I have added it back in. Thanks.

PR #63, opened 2023-05-01
FW-4285 character limit and regex fix

caforbes:
### Description of Changes

Adds error handling for a maximum character limit / alphabet size supported by custom sort (FW-4285) and renames AlphabetMapper meta to "alphabet".

Fixes a bug with regex characters in the confusable cleanup mapper, but the test for it is still broken so still skipped.

### Checklist

### Additional Notes
One fix for the broken test is to have the Alphabet query itself in order to correctly read its own json property, so there might be some issue with the test framework or testdb connection. Will make a ticket to look into a better resolution for this, it's only occurring in the test environment.


bronson-c (approved):

firstvoices/backend/models/characters.py
```diff
@@ -80,6 +82,14 @@ def validate_title_uniqueness(self):
                 % self.title
             )
 
+    def validate_character_limit(self):
+        """
+        Validates that the site has not already reached the max character limit.
+        """
+        limit = CustomSorter.max_alphabet_length
+        if len(Character.objects.filter(site_id=self.site_id)) >= limit:
```
sarahfirstvoices, line 90: Minor optimization: instead of doing len() you can add .count() to the queryset, if you don't need to retrieve all the actual rows of data.

firstvoices/backend/tests/test_models/test_alphabet_model.py
```diff
@@ -106,7 +106,6 @@ def test_clean_confusables_no_feeding(self):
         IgnoredCharacterFactory(site=alphabet.site, title="/")
         assert alphabet.clean_confusables("A/A") == "a/a"
 
-    @pytest.mark.skip("g2p with regex escape ignores mapping with escaped in-chars")
     @pytest.mark.django_db
     def test_clean_confusables_regex_escape(self):
```
sarahfirstvoices, line 110: Would it work to re-query the alphabet in the test instead? Something like...

alphabet_created = AlphabetFactory.create(....)
alphabet = Alphabet.objects.only('input_to_canonical_map').get(id=alphabet_created.id)

assert ....

firstvoices/backend/models/characters.py
```diff
@@ -256,12 +266,18 @@ def preprocess_transducer(self):
         """
         if self.input_to_canonical_map:
             preprocess_settings = self.default_g2p_config["preprocess_config"]
-
+            # need to query for self in order to correctly grab property as dict
+            self_with_object = Alphabet.objects.only("input_to_canonical_map").get(
+                id=self.id
+            )
```
sarahfirstvoices, line 272: This is a fascinating fix, wow. So the tests were having trouble handling the json correctly in some operations? If possible, it would be much better to put this fix in the test itself since it really seems like a limitation of the test framework. (Also probably worth logging a bug to the test framework project.)

caforbes: It doesn't look like it works to just put the query in the test. Seems like the test objects behave slightly differently wrt this function than actual objects do in practice. I will try to come up with a MWE and send it to the relevant test project (not sure which one exactly it comes from) -- should I keep this weird fix in, or should I go back to skipping the test?

sarahfirstvoices: Since this adds an extra database hit in a function that we will likely be running in bulk, I think we should revert to skipping that test. Maybe we need to start a list of required manual tests?

PR #69, opened 2023-05-03
Fw 4005 word of the day

sahil97:
### Description of Changes
- Adds a word of the day model
- Adds a view which selects word of the day model based on pre defined criterion , and adds additional entries if a new word is chosen for the date

### Checklist

- [x] Pull the branch and test locally (Optional)

### Additional Notes
N/A


firstvoices/backend/models/dictionary.py
```diff
@@ -259,3 +263,26 @@ class Meta:
...
+
+
+class WordOfTheDay(BaseSiteContentModel):
+    """
+    Table for word-of-the-day containing word and its respective date it was chosen to be wotd.
+    """
+
+    date = models.DateField(db_index=True)
+    dictionary_entry = models.ForeignKey(
+        "DictionaryEntry", on_delete=models.CASCADE, related_name="wotd_set"
```
sahil97, line 275: Please suggest if "cascade" works for on_delete or we should switch to some other option ?

bronson-c: It seems like the correct solution to me, on delete of a dictionary entry the corresponding wotd model is also deleted.

firstvoices/backend/models/dictionary.py
```diff
@@ -259,3 +263,26 @@ class Meta:
...
+    )
+
+    class Meta:
+        verbose_name = _("Word of the day")
+        verbose_name_plural = _("Word of the day")
```
bronson-c, line 280: Would this be "Words of the day" ?

firstvoices/backend/views/dictionary_views.py
```diff
@@ -52,3 +57,116 @@ def get_queryset(self):
...
+    pagination_class = None
+
+    @staticmethod
+    def get_suitable_queryset(site_slug):
+        # Case 1. Check if there is a word assigned word-of-the-day date of today
+        today = datetime.today()
+        selected_word = WordOfTheDay.objects.filter(
+            site__slug=site_slug, date=today
+        ).first()
+        if selected_word is not None:
```
bronson-c, line 93: could this condition just be `if selected_word:` ? Same with case 3

firstvoices/backend/views/dictionary_views.py
```diff
@@ -52,3 +57,116 @@ def get_queryset(self):
...
+    """
+    Picks and returns word-of-the-day for a given site.
+    """
+
+    http_method_names = ["get"]
+    serializer_class = DictionaryEntryDetailSerializer
+    pagination_class = None
+
+    @staticmethod
+    def get_suitable_queryset(site_slug):
```
bronson-c, line 87: For clarification, are all the returned querysets of this function supposed to be of length 1? Is that why 1 new model is saved in some of the cases?

firstvoices/backend/views/dictionary_views.py
```diff
@@ -52,3 +57,116 @@ def get_queryset(self):
...
+            site__slug=site_slug,
+            type=DictionaryEntry.TypeOfDictionaryEntry.WORD,
+            exclude_from_wotd=False,
+        ).exclude(id__in=list(words_used))
+        if queryset.count() > 0:
+            selected_word = queryset.first()
+            WordOfTheDay(
+                date=today, dictionary_entry=selected_word, site=selected_word.site
+            ).save()
+            return queryset
```
bronson-c, line 114: Could this queryset potentially return more than one word as the word of the day?

firstvoices/backend/views/dictionary_views.py
```diff
@@ -52,3 +57,116 @@ def get_queryset(self):
...
+        if selected_word is not None:
+            queryset = DictionaryEntry.objects.filter(
+                id=selected_word.dictionary_entry.id
+            )
+            if queryset.count() > 0:
+                WordOfTheDay(
+                    date=today,
+                    dictionary_entry=selected_word.dictionary_entry,
+                    site=selected_word.dictionary_entry.site,
+                ).save()
```
bronson-c, line 135: Would this create an ever-growing list of WordOfTheDay models in the db? For example if "Hello" is added as a word of the day today, then a year from now there is another model on the site with "Hello" and a different date, etc.

firstvoices/backend/admin/dictionary_admin.py
```diff
@@ -77,6 +78,14 @@ class DictionaryEntryLinkInline(BaseInlineAdmin):
...
 
 
+class WordOfTheDayInline(BaseInlineAdmin):
+    model = WordOfTheDay
+    fields = (
+        "dictionary_entry",
+        "date",
+    ) + BaseInlineAdmin.fields
+
+
```
sarahfirstvoices, line 88: Oh nice! So in a pinch, a superadmin could add a word of the day for someone on a specific date, right?

firstvoices/backend/models/dictionary.py
sarahfirstvoices: Agree

firstvoices/backend/models/dictionary.py
sarahfirstvoices: +1 for Words of the day

firstvoices/backend/models/dictionary.py
```diff
@@ -259,3 +263,26 @@ class Meta:
...
+        unique_together = ("site", "date")
+        ordering = ["-date"]
+        rules_permissions = {
+            "view": rules.always_allow,
```
sarahfirstvoices, line 284: I think the view permission should be predicates.has_visible_site

firstvoices/backend/tests/test_apis/test_wotd_api.py
```diff
@@ -0,0 +1,165 @@
...
+from backend.tests.test_apis.base_api_test import BaseSiteControlledContentApiTest
+
+
+class WordOfTheDayFactory(DjangoModelFactory):
+    dictionary_entry = factory.SubFactory(DictionaryEntryFactory)
+    date = datetime.today()
+
+    class Meta:
+        model = WordOfTheDay
+
```
sarahfirstvoices, line 27: OK I know we *just* decided to put factories right in the relevant tests, but could you please put it in the factories.py instead? It will make it work with the new list permission tests. Sorry for the back and forth!

firstvoices/backend/views/dictionary_views.py
sarahfirstvoices, line 87: This is a really long function. Can you please split it up into some smaller ones? Feel free to call me if you want to chat about it

firstvoices/backend/views/dictionary_views.py
sarahfirstvoices: Yes, the goal is a log of them :)

firstvoices/backend/views/dictionary_views.py
sarahfirstvoices: Also I think this get_or_create_todays_wotd logic might make sense to put on the model rather than the view.

firstvoices/backend/views/dictionary_views.py
```diff
@@ -52,3 +57,116 @@ def get_queryset(self):
...
+
+    def list(self, request, *args, **kwargs):
+        # Overriding list method from FVPermissionViewSetMixin to only get the first word
+        site = self.get_validated_site()
+
+        # Logic to select queryset
+        if site.count() > 0:
+            queryset = self.get_suitable_queryset(site[0].slug)
+        else:
+            queryset = DictionaryEntry.objects.none()
```
sarahfirstvoices, line 162: I don't think this logic is quite right-- get_validated_site will always return a site. (If it can't, it raises an error.)

firstvoices/backend/views/dictionary_views.py
```diff
@@ -52,3 +57,116 @@ def get_queryset(self):
...
+    SiteContentViewSetMixin,
+    mixins.ListModelMixin,
+    viewsets.GenericViewSet,
+):
+    """
+    Picks and returns word-of-the-day for a given site.
+    """
+
+    http_method_names = ["get"]
+    serializer_class = DictionaryEntryDetailSerializer
```
sarahfirstvoices, line 83: I'm thinking about future enhancements to allow writing words of the day, and I think we should update this to use a serializer that is specific to words of the day. It only needs a date field and the word.

firstvoices/backend/views/dictionary_views.py
sarahfirstvoices: It could! I don't mind either way

firstvoices/backend/admin/dictionary_admin.py
sahil97: Yes, till we have a write API for users to add a word of the day, if a community wants something specific on a date, superadmins can do that.

firstvoices/backend/tests/test_apis/test_wotd_api.py
sahil97: No worries. Thank you for adding tests that go through all the models.

firstvoices/backend/views/dictionary_views.py
sahil97: Yes, I went with the function definition of returning a queryset here. The first word is selected and then sent back in the response in the `list` function below.
But that may change after discussion with Sarah about refactoring the get_queryset function.

sahil97:
This PR also contains addition of ordering field to a couple of models to resolve pagination warnings.

sahil97:
Also, reverted a dependabot change for urllib3 library which was causing dependency conflict.

firstvoices/backend/models/part_of_speech.py
```diff
@@ -34,6 +34,7 @@ class PartOfSpeech(BaseModel):
     class Meta:
         verbose_name = _("Part Of Speech")
         verbose_name_plural = _("Parts Of Speech")
+        ordering = ["title"]
```
sarahfirstvoices, line 37: Any reason for adding all these ordering settings? They add a bit of overhead to every query

firstvoices/backend/serializers/word_of_the_day_serializers.py
```diff
@@ -0,0 +1,12 @@
...
+from backend.models.dictionary import WordOfTheDay
+from backend.serializers.dictionary_serializers import DictionaryEntryDetailSerializer
+
+
+class WordOfTheDayListSerializer(serializers.ModelSerializer):
+    dictionary_entry = DictionaryEntryDetailSerializer()
+
+    class Meta:
+        model = WordOfTheDay
+        fields = ["date", "dictionary_entry"]
```
sarahfirstvoices, line 12: What is the default date format here? Just curious

firstvoices/backend/models/part_of_speech.py
sahil97: I added these to resolve a couple of pagination warnings being shown when running pytests. These can be removed if they add a bit of overhead.

firstvoices/backend/views/word_of_the_day_views.py
```diff
@@ -0,0 +1,136 @@
...
+        random_word_id = choice(primary_keys_list)
+        random_word = DictionaryEntry.objects.get(id=random_word_id)
+        wotd_entry = WordOfTheDay(
+            date=today, dictionary_entry=random_word, site=random_word.site
+        )
+        wotd_entry.save()
+        return WordOfTheDay.objects.filter(id=wotd_entry.id)
+
+    def get_selected_word(self, site_slug):
+        # Goes over a few conditions to find a suitable word of the day
```
sarahfirstvoices, line 104: This is great! Thanks for breaking it out this way. Much easier to read

sarahfirstvoices (approved):

firstvoices/backend/serializers/word_of_the_day_serializers.py
sahil97: the default datetime field is used, but on django-admin and in the API response, it is YYYY-MM-DD

PR #71, opened 2023-05-04
Fw-4255, permissions for db queries

sarahfirstvoices:
### Description of Changes
Implements permission filters for querysets, using Q objects. I added an equivalent function for each predicate, that returns a Q object with the same logic. (Both sets are grouped in the new 'permissions' package.)

The test_permission_settings suite verifies that every predicate we use has a matching query filter, and that they return the same results. It also tests that every model has permissions configured.

Bonus changes:
* configured permissions for models that didn't yet have them
* 

### Checklist

### Additional Notes
N/A


firstvoices/backend/models/managers.py
```diff
@@ -1,17 +0,0 @@
-from django.db import models
```
sarahfirstvoices, line 1: I moved this to the permissions folder, mainly so I could add a README in there that explains how the different parts work together

sahil97 (approved):
Looking good. +1 for great parameterization in the tests

PR #73, opened 2023-05-05
Fw 4315 implement flags for read api for category

sahil97:
### Description of Changes
Adds a `contains` flag to the `category` api.

### Checklist

- [x] Pull the branch and test locally - To test locally, try creating categories, both parent and children, then assigning some of them words, some of them phrases, and some both. When hitting the api with different flags, verify that the correct categories are being returned. (Optional)

### Additional Notes
N/A


firstvoices/backend/views/category_views.py
NolanVH, line 1: During local testing, I found a case that appears to be returning duplicated categories.

If there are two dictionary entries, say `entryA` and `entryB`,  and two categories, say `catA` and `catB`, then if `catA` is a child of `catB`, `entryA` has `catA` as a category, and `entryB` has `catB` as a category then the contains= endpoint will return two copies of `catA`. One will be non-nested and one will be nested inside of the children of `catB`.

Duplication does not happen for the main `api/1.0/sites/<site-slug>/categories/` URL. Child categories only appear nested within a parent category `children` field.

NolanVH, line 1: It would probably be helpful to add parameters to the `@extend_schema_view` decorator to give information in the API docs.

NolanVH (changes requested):
Aside from the minor points in my comments, this change looks great! Thanks @sahil97.

firstvoices/backend/tests/test_apis/test_category_api.py
```diff
@@ -104,3 +101,155 @@ def test_detail_404(self):
...
+        response = self.client.get(
+            self.get_list_endpoint(
+                self.site.slug, query_kwargs={"contains": TypeOfDictionaryEntry.WORD}
+            )
+        )
+        response_data = json.loads(response.content)
+
+        assert response.status_code == 200
+        assert response_data["count"] == 0
+
```
sarahfirstvoices, line 176: Please add a test case for what happens on invalid input, like {url}?type=oyster

firstvoices/backend/tests/test_apis/test_category_api.py
```diff
@@ -104,3 +101,155 @@ def test_detail_404(self):
...
+            str(category_word.id),
+            str(category_phrase.id),
+            str(category_both.id),
+        ]
+
+        assert len(actual_ids) == len(ids_in_response)
+        # Checking all ids match up
+        difference = set(actual_ids) ^ set(ids_in_response)
+        assert not difference
+
```
sarahfirstvoices, line 218: Please add another test case for what happens when there are multiple choices that include an invalid option, like ?type=word|oyster

firstvoices/backend/tests/test_apis/test_category_api.py
```diff
@@ -104,3 +101,155 @@ def test_detail_404(self):
...
+        # Testing for PHRASE flag, should not return anything
+        response = self.client.get(
+            self.get_list_endpoint(
+                self.site.slug, query_kwargs={"contains": TypeOfDictionaryEntry.PHRASE}
+            )
+        )
+        response_data = json.loads(response.content)
+
+        assert response.status_code == 200
+        assert response_data["count"] == 0
```
sarahfirstvoices, line 138: URLs need to be case-insensitive, so please add a test case to show that different casings work. (You could parametrize this one, with something like ["PHRASE", "phrase", "PhRase"]

firstvoices/backend/tests/test_apis/test_category_api.py
```diff
@@ -104,3 +101,155 @@ def test_detail_404(self):
...
+            str(response_obj["id"]) for response_obj in response_data["results"]
+        ]
+        actual_ids = [
+            str(parent_category.id),
+            str(child_category.id),
+        ]
+
+        assert len(actual_ids) == len(ids_in_response)
+        # Checking all ids match up
+        difference = set(actual_ids) ^ set(ids_in_response)
```
sarahfirstvoices, line 254: I think you can just assert that the sets are equal

firstvoices/backend/tests/test_apis/test_category_api.py
```diff
@@ -104,3 +101,155 @@ def test_detail_404(self):
...
+            str(category_both.id),
+        ]
+
+        assert len(actual_ids) == len(ids_in_response)
+        # Checking all ids match up
+        difference = set(actual_ids) ^ set(ids_in_response)
+        assert not difference
+
+    @pytest.mark.django_db
+    def test_nested_categories(self):
```
sarahfirstvoices, line 220: Is this meant to test that contains also applies to the child categories? If so I think we need to add one child that is a phrase, to prove it gets filtered out. And maybe add "contains" in the test name

firstvoices/backend/views/category_views.py
```diff
@@ -44,11 +46,34 @@ def get_queryset(self):
...
-                Category.objects.filter(site__slug=site[0].slug)
-                .prefetch_related("children")
-                .exclude(parent__isnull=False)
-            )
+            list_queryset = Category.objects.filter(site__slug=site[0].slug)
+
+            # Check if type flags are present
+            contains_flags = self.request.GET.get("contains", "").split("|")
+
+            if all(
```
sarahfirstvoices, line 54: I think this would be more straightforward and also more expandable as a list of filters instead of a series of if/else cases. Maybe something like:

for each supported type
      if the type is in the contains_flags, add the necessary Q filter to a list

type_filter = reduce(operator.or_, list_of_q_objects, Q())

list_queryset.filter(type_filter)
list_queryset.filter(parent__isnull=False)

firstvoices/backend/tests/test_apis/test_category_api.py
```diff
@@ -104,3 +101,155 @@ def test_detail_404(self):
         wrong_endpoint = self.get_detail_endpoint(self.site.slug, "54321")
         response = self.client.get(wrong_endpoint)
         assert response.status_code == 404
+
+    @pytest.mark.django_db
+    def test_contains_word(self):
+        word_entry = DictionaryEntryFactory(site=self.site)
+        category_word = ParentCategoryFactory(site=self.site)
+        category_word.dictionary_entries.add(word_entry)
```
sarahfirstvoices, line 109: For all of these single-type cases, you need to also add some entries to the categories that are going to be filtered out. I don't think this test would fail if even if the code was broken

firstvoices/backend/tests/test_apis/test_category_api.py
```diff
@@ -104,3 +101,155 @@ def test_detail_404(self):
...
+        response_data = json.loads(response.content)
+
+        assert response.status_code == 200
+        assert response_data["count"] > 0
+
+        assert response_data["results"][0] == {
+            "id": str(category_word.id),
+            "title": category_word.title,
+            "description": category_word.description,
+            "children": [],
```
sarahfirstvoices, line 126: No need to test the entire data object here, since there are already tests for the data format. This test only needs to check for the correct ids

sarahfirstvoices: (Same for all of these new tests)

firstvoices/backend/views/category_views.py
sahil97: Good catch. I've just confirmed with Guy regarding this. For the MVP perspective, the duplicates will be removed so the response for `contains=` endpoint will be similar to the main api response.

NolanVH (approved):
Tested locally working as intended. The added parameters to the API docs look great. Thanks @sahil97.

bronson-c (approved):
LGTM 👍

firstvoices/backend/tests/test_apis/test_category_api.py
```diff
@@ -102,91 +115,86 @@ def test_detail_404(self):
...
             )
         )
         response_data = json.loads(response.content)
 
         assert response.status_code == 200
-        assert response_data["count"] == 0
+        assert response_data["count"] == expected_count
 
+    @pytest.mark.parametrize(
+        "flag, expected_count", [("WORD", 1), ("word", 1), ("WorD", 1)]
```
sarahfirstvoices, line 151: Oh nice, I didn't know you could declare paired parameters this way

firstvoices/backend/tests/test_apis/test_category_api.py
```diff
@@ -211,45 +219,96 @@ def test_contains_multiple(self):
             str(category_both.id),
         ]
 
-        assert len(actual_ids) == len(ids_in_response)
-        # Checking all ids match up
-        difference = set(actual_ids) ^ set(ids_in_response)
-        assert not difference
+        assert set(actual_ids) == set(ids_in_response)
```
sarahfirstvoices, line 222: Minor: could you please change "actual_ids" to "expected_ids" since it is the test expectation? It's convention to assert "expected" vs "actual" so it's confusing to use "actual" for the expectation

sarahfirstvoices (approved):

PR #77, opened 2023-05-08
Add superuser membership creation to reset script

NolanVH:
### Description of Changes
This change adds the creation of an app-level membership model to the `reset-local-database.sh` script, giving the default admin user Superuser membership.

To test run the `reset-local-database.sh` script and see that an app-level membership with the `Superadmin` role has been created for the local admin user.

### Checklist

### Additional Notes
N/A


bronson-c (approved):
Good to add but would it be preferable to put `firstvoices/create_local_superadmin_membership.py` inside a `scripts` directory? I was considering adding another script that populates the db with some test models for local testing and that addition to the reset script could also go there.

sahil97 (approved):
Tested locally. Works as expected.

sarahfirstvoices (approved):

NolanVH:
> Good to add but would it be preferable to put `firstvoices/create_local_superadmin_membership.py` inside a `scripts` directory? I was considering adding another script that populates the db with some test models for local testing and that addition to the reset script could also go there.

Good idea @bronson-c. I have moved it into a `scripts` directory.

PR #104, opened 2023-05-12
FW-4355 Recalculate Custom Order Preview

bronson-c:
### Description of Changes
This PR adds and endpoint to view a custom order recalculation preview, without actually running the recalculation on the db.

Sending a POST request to `api/1.0/sites/<site-slug>/dictionary-cleanup/preview` will start a task to generate the recalculation preview which runs asynchronously and saves the results in their own model.

Sending a GET request to the same url will return the most recently generated preview result, along with the date of it's generation and the status of any ongoing recalculate preview tasks.

### Checklist

### Additional Notes

The preview results are in the following format:
```
{
    "results": [
        {
            "site": "Test Site",
            "currentPreviewTaskStatus": "SUCCESS",
            "latestRecalculationDate": "2023-05-26T11:37:20.413913-07:00",
            "latestRecalculationResult": {
                "unknownCharacterCount": {},
                "updatedEntries": [
                    {
                        "title": "test",
                        "cleanedTitle": "",
                        "isTitleUpdated": false,
                        "previousCustomOrder": "⚑t⚑e⚑s⚑t",
                        "newCustomOrder": "!$%!"
                    },
                    {
                        "title": "ᐱᐱᐱ",
                        "cleanedTitle": "AAA",
                        "isTitleUpdated": true,
                        "previousCustomOrder": "⚑ᐱ⚑ᐱ⚑ᐱ",
                        "newCustomOrder": "###"
                    }
                ]
            }
        }
    ]
}
```


firstvoices/backend/rules.py
bronson-c, line 1: In order for this rule to register it must be in a file named `rules.py` in this location. However it may be possible to add a rule inside a model if desired.

firstvoices/backend/tests/test_async/test_async.py
```diff
@@ -0,0 +1,129 @@
...
+        response = self.client.get(
+            f"{self.get_detail_endpoint(site.slug)}dictionary-cleanup/preview"
+        )
+        response_data = json.loads(response.content)
+        assert response.status_code == 404
+        assert response_data == {
+            "message": "No recalculation results found for this site."
+        }
+
+    @pytest.mark.django_db(transaction=True, serialized_rollback=True)
```
bronson-c, line 78: In order for async tests to run sequentially in our test db, `serialized_rollback` must be enabled. This does slow the test down however.

firstvoices/firstvoices/settings_test.py
```diff
@@ -5,7 +5,7 @@
 from .settings import *  # noqa
 
 # https://docs.djangoproject.com/en/dev/ref/settings/#test-runner
-TEST_RUNNER = "django.test.runner.DiscoverRunner"
+TEST_RUNNER = "djcelery.contrib.test_runner.CeleryTestSuiteRunner"
```
bronson-c, line 8: This enables `CELERY_ALWAYS_EAGER` and celery fixtures by default.

firstvoices/pytest.ini
```diff
@@ -1,3 +1,3 @@
 [pytest]
-addopts = --ds=firstvoices.settings_test --reuse-db
+addopts = --ds=firstvoices.settings_test --create-db
```
bronson-c, line 2: In order to test async functions, we must re-create the db on test run to preserve fixtures.

requirements.txt
```diff
@@ -63,6 +63,7 @@ pycparser==2.21
 PyJWT==2.7.0
 pyrsistent==0.19.3
 pytest==7.3.1
+pytest-celery==0.0.0
```
bronson-c, line 66: This requirement actually just works as a flag to enable `pytest-celery`

firstvoices/backend/views/custom_order_recalculate_views.py
```diff
@@ -0,0 +1,81 @@
...
+                (site_slug,)
+            )
+            self.task_id = recalculation_results.task_id
+
+            CustomOrderRecalculationPreviewResult.objects.create(
+                site=site, result=recalculation_results.get(timeout=360)
+            )
+
+            return Response(
+                {"message": "Successfully saved recalculation results"}, status=201
```
sarahfirstvoices, line 74: Since this is returned as soon as the task is registered, this message should probably be more like, "Recalculation is queued."

firstvoices/backend/views/custom_order_recalculate_views.py
```diff
@@ -0,0 +1,81 @@
...
+            )
+
+        except recalculate_custom_order_preview.OperationalError:
+            return Response(
+                {"message": "An error occurred while dispatching remote task"},
+                status=503,
```
sarahfirstvoices, line 80: I think a 500 makes more sense here. 503 usually means the server is too busy or your request got throttled

firstvoices/pytest.ini
sarahfirstvoices: This solves migration conflicts during local development as well-- good change

firstvoices/backend/views/custom_order_recalculate_views.py
```diff
@@ -0,0 +1,81 @@
...
+        # If there is no result and no task_id, then there are no recalculation results for this site
+        if not self.task_id and not result:
+            return Response(
+                {"message": "No recalculation results found for this site."}, status=404
+            )
+
+        # If there is a task_id, check the status of the task and add it to the response
+        if self.task_id:
+            async_result = AsyncResult(self.task_id)
+            preview_info = {"preview_current_task_status": async_result.status}
```
sarahfirstvoices, line 38: Please put these json keys in variables if they are used in more than one place. Usually my rule is two repeats is ok, but for something going in a public response it needs to be more rigorous.

sarahfirstvoices: (Same comment for other keys below.)

firstvoices/backend/views/custom_order_recalculate_views.py
```diff
@@ -0,0 +1,81 @@
...
+
+        result = (
+            CustomOrderRecalculationPreviewResult.objects.filter(site__slug=site_slug)
+            .order_by("-date")
+            .first()
+        )
+
+        # If there is no result and no task_id, then there are no recalculation results for this site
+        if not self.task_id and not result:
+            return Response(
```
sarahfirstvoices, line 31: Would it make sense here to return the same "Not started" status as below instead of the 404? Since this is a status report, I think it could make sense to always return a result, and sometimes the result is just "nothing to see here"

sarahfirstvoices: I think if you skipped this conditional, the code below would still work and make sense for this case

firstvoices/backend/views/custom_order_recalculate_views.py
```diff
@@ -0,0 +1,81 @@
...
+from backend.models import CustomOrderRecalculationPreviewResult, Site
+from backend.tasks.alphabet_tasks import recalculate_custom_order_preview
+
+
+class CustomOrderRecalculatePreviewView(APIView):
+    task_id = None
+
+    @staticmethod
+    def check_superadmin_status(request):
+        if not rules.has_perm("views.has_custom_order_access", request.user):
```
sarahfirstvoices, line 16: This is checking a permission, so please call it something like check_permission

firstvoices/backend/views/custom_order_recalculate_views.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def post(self, request, site_slug: str):
+        # Check Superadmin status
+        self.check_superadmin_status(request)
+
+        # Check that the site exists
+        try:
+            site = Site.objects.get(slug=site_slug)
+        except Site.DoesNotExist:
+            return Response({"message": "Site not found"}, status=404)
```
sarahfirstvoices, line 60: Instead of this, please use `raise Http404` so we get the default handling

firstvoices/backend/views/custom_order_recalculate_views.py
sarahfirstvoices: Also, please raise an exception instead of manually constructing an error response. (If we change our default error response format in future, we don't want to have to separately edit this one.) It's probably worth making a custom APIException subclass for this. E.g., https://www.django-rest-framework.org/api-guide/exceptions/#apiexception

firstvoices/backend/tasks/alphabet_tasks.py
```diff
@@ -0,0 +1,55 @@
...
+        cleaned_title = alphabet.clean_confusables(entry.title)
+        new_order = alphabet.get_custom_order(cleaned_title)
+
+        result["cleaned_title"] = cleaned_title
+        result["new_custom_order"] = new_order
+
+        if "⚑" in new_order:
+            entries_with_unknown_chars.append(entry)
+
+        updated_entries.append(result)
```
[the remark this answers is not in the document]
sarahfirstvoices: Oh I see, it iterates them below. Better to just not add it in first place, I think.

firstvoices/backend/tasks/alphabet_tasks.py
```diff
@@ -0,0 +1,55 @@
...
+    # Then get the count of unknown characters
+    unknown_character_count = {}
+    for entry in entries_with_unknown_chars:
+        chars = alphabet.get_character_list(entry.title)
+        for char in chars:
+            custom_order = alphabet.get_custom_order(char)
+            if "⚑" in custom_order:
+                if custom_order not in unknown_character_count:
+                    unknown_character_count[custom_order] = 0
+                unknown_character_count[custom_order] += 1
```
sarahfirstvoices, line 50: Since we are dealing with such potentially large lists, it would be better to update these counts during the main iteration instead of iterating again.

firstvoices/backend/views/custom_order_recalculate_views.py
```diff
@@ -0,0 +1,81 @@
...
+            raise PermissionDenied()
+
+    def get(self, request, site_slug: str):
+        # Check Superadmin status
+        self.check_superadmin_status(request)
+
+        result = (
+            CustomOrderRecalculationPreviewResult.objects.filter(site__slug=site_slug)
+            .order_by("-date")
+            .first()
```
sarahfirstvoices, line 26: Is there any reason you couldn't use this as the queryset for the view? Then I think you could use the built-in permission checking based on the model permissions

firstvoices/backend/tests/test_async/test_alphabet_tasks.py
```diff
@@ -0,0 +1,100 @@
...
+
+class TestAlphabetTasks:
+    @pytest.fixture
+    def site(self):
+        return factories.SiteFactory.create(slug="test")
+
+    @pytest.fixture
+    def alphabet(self, site):
+        return factories.AlphabetFactory.create(site=site)
+
```
sarahfirstvoices, line 15: Please add a test case for getting a preview when none has been calculated

firstvoices/backend/tests/test_async/test_alphabet_tasks.py
```diff
@@ -0,0 +1,100 @@
...
+                    "new_custom_order": "#$%⚑d",
+                    "previous_custom_order": "⚑a⚑b⚑c⚑d",
+                },
+                {
+                    "title": "ᐱᐱᐱ",
+                    "cleaned_title": "AAA",
+                    "new_custom_order": "!!!",
+                    "previous_custom_order": "⚑ᐱ⚑ᐱ⚑ᐱ",
+                },
+            ],
```
sarahfirstvoices, line 99: I think we need a couple more test cases:

- a dictionary entry that had no unknown characters, and is unaffected by the recalculate
- a dictionary entry that does have unknown characters, and is unaffected by the recalculate
- a dictionary entry that had no unknown characters, and is changed by the recalculate

firstvoices/backend/tests/test_async/test_async.py
sarahfirstvoices: Can we adjust the test so they don't need to run sequentially? Sequential tests usually cause multiple problems down the line

firstvoices/backend/tests/test_async/test_async.py
```diff
@@ -0,0 +1,129 @@
...
+
+import pytest
+
+from backend.models.constants import AppRole, Visibility
+from backend.tests import factories
+from backend.tests.test_apis.base_api_test import BaseApiTest
+
+
+class TestCustomOrderRecalculatePreview(BaseApiTest):
+    API_DETAIL_VIEW = "api:site-detail"
```
sarahfirstvoices, line 12: This belongs in the test_api folder in a file called test_dictionary_cleanup_api

firstvoices/backend/tests/test_async/test_alphabet_tasks.py
sarahfirstvoices: Oops, found it :)

firstvoices/backend/tests/test_async/test_async.py
```diff
@@ -0,0 +1,129 @@
...
+from unittest.mock import ANY
+
+import pytest
+
+from backend.models.constants import AppRole, Visibility
+from backend.tests import factories
+from backend.tests.test_apis.base_api_test import BaseApiTest
+
+
+class TestCustomOrderRecalculatePreview(BaseApiTest):
```
sarahfirstvoices, line 11: Since this is testing the api, I think it could be named after the api. Either, TestDictionaryCleanupPreview or just TestPreview

firstvoices/backend/tests/test_async/test_async.py
sarahfirstvoices: Would it be as simple as using a new site for each test and creating the necessary data for the test case? That would make the tests easier to understand as well

firstvoices/backend/tests/test_async/test_async.py
```diff
@@ -0,0 +1,129 @@
...
+        self, celery_worker, site, alphabet, django_db_serialized_rollback
+    ):
+        user = factories.get_app_admin(role=AppRole.SUPERADMIN)
+        self.client.force_authenticate(user=user)
+        assert user.has_perm(self.SUPERADMIN_PERMISSION, site)
+
+        factories.DictionaryEntryFactory.create(site=site, title="test")
+
+        response_post = self.client.post(
+            f"{self.get_detail_endpoint(site.slug)}dictionary-cleanup/preview"
```
sarahfirstvoices, line 107: Seems like it would be worth overriding get_detail_endpoint to return this whole url

sarahfirstvoices (changes requested):
Bronson this is looking so good! Msg me if you want to chat about any of my comments

firstvoices/backend/views/custom_order_recalculate_views.py
bronson-c: I've changed this to raise the built in drf server error (500) but if you'd prefer a custom APIException I can make one

firstvoices/backend/views/custom_order_recalculate_views.py
```diff
@@ -0,0 +1,81 @@
...
+
+        # Call the recalculation preview task
+        try:
+            recalculation_results = recalculate_custom_order_preview.apply_async(
+                (site_slug,)
+            )
+            self.task_id = recalculation_results.task_id
+
+            CustomOrderRecalculationPreviewResult.objects.create(
+                site=site, result=recalculation_results.get(timeout=360)
```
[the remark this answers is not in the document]
bronson-c: Good suggestion. I have set the previous result models to be deleted when a new preview for that site is requested

firstvoices/backend/tasks/alphabet_tasks.py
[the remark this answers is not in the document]
bronson-c: Great suggestion, I have added it. However @caforbes this means if a character's cleaned title it the same as the title, the result will be:
```
"updated_entries": [
                {
                    "title": "abc",
                    "cleaned_title": "",
                    "new_custom_order": "!#$",
                    "previous_custom_order": "⚑a⚑b⚑c",
                }
            ],
```
Is that acceptable or would it be preferable to have an additional bit of information to say the cleaned title is the same as title?

firstvoices/backend/tests/test_async/test_async.py
bronson-c: I'll give it a go, but serialized rollback may be required as after any test with transaction=true the default g2p config will be removed.

firstvoices/backend/tasks/alphabet_tasks.py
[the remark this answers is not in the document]
caforbes: In my head that was what I wanted, but you are right that it's probably potentially confusing and there should be a little more info. Maybe a more informative solution could be to have a flag `"is_title_updated": true/false`, and then have the updated title under `cleaned_title`?

firstvoices/backend/views/custom_order_recalculate_views.py
```diff
@@ -0,0 +1,53 @@
...
+from backend.serializers.async_results_serializers import (
+    CustomOrderRecalculationPreviewResultDetailSerializer,
+)
+from backend.tasks.alphabet_tasks import recalculate_custom_order_preview
+from backend.views.base_views import FVPermissionViewSetMixin, SiteContentViewSetMixin
+from backend.views.exceptions import CeleryError
+
+
+class CustomOrderRecalculatePreviewView(
+    FVPermissionViewSetMixin, SiteContentViewSetMixin, ModelViewSet
```
NolanVH, line 14: I noticed that the endpoint `/api/1.0/sites/{site_slug}/dictionary-cleanup/preview/{id}/` is still enabled and shown in the `/api/docs/` page. It looks like this endpoint is not needed and could be removed. An easy way to do that is to subclass the view with the needed subset of mixins in `ModelViewSet` (I believe it is `mixins.RetrieveModelMixin` that adds the `/{id}/` endpoint).

NolanVH (approved):
This is looking awesome! Thanks @bronson-c.

Tested locally working as intended. All changes look good to me and my one additional comment is very minor and not mandatory so I will approve now.

firstvoices/backend/serializers/async_results_serializers.py
```diff
@@ -0,0 +1,51 @@
...
+        async_result = AsyncResult(obj.task_id)
+        return async_result.status
+
+    def get_latest_recalculation_result(self, obj):
+        ordered_result = OrderedDict(
+            {
+                "unknown_character_count": obj.latest_recalculation_result[
+                    "unknown_character_count"
+                ],
+                "updated_entries": [],
```
sarahfirstvoices, line 26: Maybe I'm missing something. Can this just be `"updated_entries": obj.latest_recalculation_result["updated_entries"]` and skip the transcribing below?

firstvoices/backend/tests/test_tasks/test_alphabet_tasks.py
```diff
@@ -0,0 +1,154 @@
...
+            "unknown_character_count": {},
+            "updated_entries": [
+                {
+                    "title": "aab",
+                    "is_title_updated": False,
+                    "cleaned_title": "",
+                    "previous_custom_order": "!!#",
+                    "new_custom_order": "$#",
+                }
+            ],
```
sarahfirstvoices, line 153: These are great. Thanks for the new tests :)

sarahfirstvoices (approved):

firstvoices/backend/views/base_views.py
```diff
@@ -101,3 +103,17 @@ def get_serializer_context(self):
...
         return context
+
+
+class ListViewOnlyModelViewSet(
+    GenericViewSet,
+    mixins.CreateModelMixin,
+    mixins.UpdateModelMixin,
+    mixins.DestroyModelMixin,
+    mixins.ListModelMixin,
+):
```
sarahfirstvoices, line 114: Minor: put the mixins before the main base class please

firstvoices/backend/views/custom_order_recalculate_views.py
bronson-c: Nice catch! Made a small change to disable the detail view

firstvoices/backend/serializers/async_results_serializers.py
bronson-c: The reason I'm transcribing is just for the proper order. The second OrderedDict is there to ensure the order of the "updated entry" fields are consistent. They sometimes get jumbled around in the JSONfield so the two ordered dicts ensure consistency with {"unknown_character_count", "updated_entries"} and {"title", "cleaned_title", is_title_updated", "previous_custom_order", "new_custom_order"} respectively

PR #118, opened 2023-05-17
FW-4394 Reduce duplicate queries in dictionary entry views

NolanVH:
### Description of Changes
These changes add a context object to the dictionary views which is then passed to the dictionary entry serializer. This context object contains alphabet, ignored character, base character, character variant, and ignorable character information. This information can be collected once and reused during the fetching of each dictionary entry instead of re-querying the database each time.

If testing locally, please try the dictionary list view and detail views and compare the number of queries before and after the changes. The number should at least stay the same or decrease significantly. The changes are more noticeable on a site with a larger number of characters, character variants, ignored characters, and dictionary entries. The number of queries can be seen by using the toolbar in debug mode.

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/models/characters.py
+++ b/firstvoices/backend/models/characters.py
@@ -221,6 +221,27 @@ class Alphabet(BaseSiteContentModel):
     input_to_canonical_map = models.JSONField(default=list)
 
     @property
+    def base_characters(self):
+        """
+        Characters for the site in sort order.
+        """
+        return Character.objects.filter(site=self.site).order_by("sort_order")
+
+    @property
+    def variant_characters(self):
+        """
+        All variant characters for the site.
+        """
+        return CharacterVariant.objects.filter(site=self.site)
+
+    @property
+    def ignorable_characters(self):
+        """
+        Ignorable characters for the site.
+        """
+        return IgnoredCharacter.objects.filter(site=self.site)
+
+    @property
     def default_g2p_config(self):
         """
         Returns default G2P configurations for required mappers, customized with site name and slug.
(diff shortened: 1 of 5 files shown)


firstvoices/backend/models/characters.py
```diff
@@ -277,47 +256,77 @@ def preprocess_transducer(self):
             self.logger.warning("Empty confusable map for site %s", self.site)
             return None
 
-    @property
-    def presort_transducer(self):
+    def presort_transducer(self, context):
```
sarahfirstvoices, line 259: Instead of passing a dictionary here, let's use named parameters for each specific thing and give them defaults of None. So something like 

def presort_transducer(self, base_characters=None, character_variants=None, etc)

firstvoices/backend/models/characters.py
```diff
@@ -220,27 +220,6 @@ class Meta:
...
-        """
-        return CharacterVariant.objects.filter(site=self.site)
-
-    @property
-    def ignorable_characters(self):
-        """
-        Ignorable characters for the site.
-        """
-        return IgnoredCharacter.objects.filter(site=self.site)
-
```
sarahfirstvoices, line 243: Any reason to remove these properties? Seems like they could still be a useful way to get the default values in the other functions, if the context/parameters are not missing

firstvoices/backend/models/characters.py
sarahfirstvoices: Same for the rest of the functions as well

firstvoices/backend/views/base_views.py
```diff
@@ -42,9 +42,7 @@ def list(self, request, *args, **kwargs):
         # paginated response
         page = self.paginate_queryset(queryset)
         if page is not None:
-            serializer = self.get_serializer(
-                page, many=True, context={"request": request}
```
sarahfirstvoices, line 46: Good catch, here

sarahfirstvoices (changes requested):

sarahfirstvoices:
This definitely reduces the db queries I'm seeing locally (dictionary list went from 105 down to 42) :)

firstvoices/backend/models/characters.py
bronson-c: I have the same question, do they have to be removed to reduce the queries?

bronson-c (commented):
This is really slick! I share the opinions Sarah has for any changes but one those are added I'll add my approval. Great stuff

firstvoices/backend/serializers/dictionary_serializers.py
```diff
@@ -73,53 +75,86 @@ class DictionaryEntryDetailSerializer(serializers.HyperlinkedModelSerializer):
...
 
-    @staticmethod
-    def get_split_chars(entry):
-        alphabet = Alphabet.objects.filter(site=entry.site).first()
-        ignored_characters = alphabet.ignorable_characters.values_list(
-            "title", flat=True
-        )
+    logger = logging.getLogger(__name__)
+
+    def retrieve_model_or_context(self, model_name):
```
sarahfirstvoices, line 80: Nitpick: this method name is not really accurate anymore. Maybe rename to get_model_from_context ?

sarahfirstvoices (approved):

firstvoices/backend/serializers/dictionary_serializers.py
NolanVH: Good catch, thanks. I have updated the function name.

bronson-c (approved):

PR #121, opened 2023-05-17
fw-4409, eliminate db lookup for site content hyperlink fields

sarahfirstvoices:
### Description of Changes
Straight refactoring with no visible change in the responses. The SiteHyperlinkedRelatedField and Identity field were making one db request every time they generated a url, in order to get the site slug. This would be reasonable if we were linking to objects from multiple sites, but for all of our current cases we are always linking within the same site. I changed it to reuse the site slug from the view’s kwargs, to eliminate the extra db queries. If multi-site cases come up in the future, we can use NestedHyperlinkedRelatedField.

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/serializers/fields.py
+++ b/firstvoices/backend/serializers/fields.py
@@ -1,12 +1,58 @@
-from rest_framework_nested.relations import (
-    NestedHyperlinkedIdentityField,
-    NestedHyperlinkedRelatedField,
-)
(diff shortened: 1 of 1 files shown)


NolanVH (approved):
Nice catch on the optimization! All changes look good to me. Thanks, @sarahfirstvoices!

PR #122, opened 2023-05-17
Fw 4378 all site search api

sahil97:
### Description of Changes
Addition of a basic view to implement site-wide search with just two parameters as of now.
This PR contains basic setup for elastic search addition to the backend such as indexing, documents and one view with one parameter.
[realted wiki page](https://firstvoices.atlassian.net/wiki/spaces/FIR/pages/275251209/ElasticSearch+Draft

### Checklist

- [x] Pull the branch and test locally (Refer to instructions below for local testing) 

### Additional Notes

To test locally:

1. Ensure that elastic-search server is running locally. Instructions can be found in the readme to get the docker container running.
2. Add any new words and they will be indexed automatically. Updating or deleting should also be tested.
3. The following url can be used with the q parameter
url: http://localhost:8000/api/1.0/search/?q=wordTitle
```
q: enter a search query which can be matched with the title field in the dictionaryEntry model as of now.
```
4. Verify that the search results returned contains the query param.


NolanVH (changes requested):
This is looking great aside from the few minor issues in my comments! Thanks @sahil97! 

One thing to note, which may be intentional, is that if words titled `materials` and `mat` both exist and a search using the query `/api/1.0/search/?q=mat` is performed, then `mat` entry will be returned, but not the `materials` entry.

It would also be good to add some unit tests for basic search functionality.

Additionally, the CI test workflow is failing due to elastic-search not being available in the CI runs. We should probably add that to this PR and I would be happy to connect to set that up if you would like.

NolanVH (approved):
Given the plan to add wildcard matching, permissions checking, tests, and pagination in future PRs, these changes all look good to me. Tested locally and working as intended. Thanks @sahil97!

firstvoices/backend/views/search/utils.py
```diff
@@ -0,0 +1,45 @@
+from backend.models.dictionary import DictionaryEntry
```
sarahfirstvoices, line 1: Minor, but I think this could go in /search instead of /views/search

firstvoices/backend/views/search/search_query_builder.py
```diff
@@ -0,0 +1,47 @@
+from elasticsearch import Elasticsearch
```
sarahfirstvoices, line 1: Minor, but I think this could go in /search instead of /views/search

firstvoices/backend/views/search/custom_search_views.py
```diff
@@ -0,0 +1,45 @@
...
+        Function to build and execute the search query.
+        Returns raw objects returned from elastic-search.
+        """
+        search_params = self.get_search_params()
+        search_query = get_search_query(q=search_params["q"])
+        raw_objects = []
+        response = search_query.execute()
+        if response["hits"]["total"]["value"]:
+            for hit in response["hits"]["hits"]:
+                raw_objects.append(hit)
```
sarahfirstvoices, line 31: Is there are reason not to just return response["hits"]["hits"] ?

firstvoices/backend/views/search/custom_search_views.py
```diff
@@ -0,0 +1,45 @@
...
+    def get_search_params(self):
+        """
+        Function to return search params in a structured format.
+        """
+        input_q = self.request.GET.get("q", "")
+
+        return {"q": input_q}
+
+    def get_raw_objects(self):
+        """
```
sarahfirstvoices, line 21: Since this is actually executing the search, I think it deserves a more descriptive name. Maybe search(), or get_search_results() or get_search_hits()

firstvoices/backend/views/search/utils.py
```diff
@@ -0,0 +1,45 @@
...
+from backend.search_indices.dictionary_entry_document import (
+    ELASTICSEARCH_DICTIONARY_ENTRY_INDEX,
+)
+
+
+def hydrate_objects(raw_objects):
```
sarahfirstvoices, line 7: Is raw_objects a technical term? Could we call them whatever elasticsearch calls them? Hits or search_results or whatever?

firstvoices/backend/views/search/utils.py
```diff
@@ -0,0 +1,45 @@
...
+                if len(translation_entries):
+                    for translation in translation_entries:
+                        translations.append(
+                            {"id": translation.id, "text": translation.text}
+                        )
+                complete_object["translations"] = translations
+
+        complete_objects.append(complete_object)
+
+    return complete_objects
```
sarahfirstvoices, line 45: I think this hydration process is going to need some work, especially once we are supporting more than one datatype. My first concern is that it would be great to not make a separate db request for every object. Second concern is that it would be great if we could reuse our existing serializers for the results instead of creating this new way of producing json.

sarahfirstvoices (changes requested):

firstvoices/backend/views/search/custom_search_views.py
sahil97: You're right.  I was doing some manipulation here before, looks like I moved the manipulation but left the loop here. Good catch. Thanks

firstvoices/backend/views/search/custom_search_views.py
sahil97: After removing the for loop above, there wasn't much going on in the function except for executing the query and returning the hits. I thus moved the two lines into the list function.

firstvoices/backend/search/indices/dictionary_entry_document.py
```diff
@@ -0,0 +1,60 @@
...
+    # Add document to es index
+    try:
+        index_entry = DictionaryEntryDocument(
+            _id=instance.id, title=instance.title, type=instance.type
+        )
+        index_entry.save()
+    except ConnectionError:
+        logger = logging.getLogger(ELASTICSEARCH_LOGGER)
+        logger.warning(
+            "Elasticsearch server down. Document could not be updated in index."
```
sarahfirstvoices, line 43: I think it would be helpful to include the instance id in these warning messages

firstvoices/backend/search/indices/dictionary_entry_document.py
```diff
@@ -0,0 +1,60 @@
...
+
+    try:
+        index_entry = DictionaryEntryDocument.get(id=instance.id)
+        index_entry.delete()
+    except ConnectionError:
+        logger.warning(
+            "Elasticsearch server down. Documents could not be deleted in index."
+        )
+    except NotFoundError:
+        logger.warning("Indexed document not found. Cannot delete from index.")
```
sarahfirstvoices, line 60: Thanks for adding these signals! I think it's much nicer to keep all this together. Someday we might want to make the signal handling generic, similar to how django-elasticsearch-dsl works, but this is great for this ticket.

firstvoices/backend/search/utils/object_utils.py
```diff
@@ -0,0 +1,54 @@
...
+        DictionaryEntry.objects.filter(
+            id__in=dictionary_search_results_ids
+        ).prefetch_related("translation_set")
+    )
+
+    for obj in search_results:
+        # Handling DictionaryEntry objects
+        if obj["_index"] == ELASTICSEARCH_DICTIONARY_ENTRY_INDEX:
+            dictionary_entry = get_object(dictionary_objects, obj["_id"])
+            request.parser_context["kwargs"]["site_slug"] = dictionary_entry.site.slug
```
sarahfirstvoices, line 44: This is reminding me that I think we should modify our views/serializers to put the site_slug as a top-level property in the context, rather than having to hunt around for it in the request and/or view. No need to do that for this ticket tho-- I'll add an issue just to make that change

firstvoices/backend/search/utils/object_utils.py
```diff
@@ -0,0 +1,54 @@
...
+
+            # Serializing and adding the object to complete_objects
+            complete_objects.append(
+                DictionaryEntryDetailSerializer(
```
sarahfirstvoices, line 48: Do we need to include some of the search result info here, rather than just a straight dictionary entry detail response? I think it might be helpful to have a DictionaryEntrySearchResultSerializer that is returns something like:

`{
search_property_1: value,
search_property_2: value,
result: <DictionaryDetailResponse>
}`

firstvoices/backend/tests/test_search/test_query_params.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+
+from backend.views.search.search_query_builder import get_search_query
+
+
```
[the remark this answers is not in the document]
sarahfirstvoices: Marking this resolved because it can be handled in FW-4433

firstvoices/backend/views/search/custom_search_views.py
```diff
@@ -17,27 +19,24 @@ def get_search_params(self):
 
         return {"q": input_q}
 
-    def get_raw_objects(self):
-        """
-        Function to build and execute the search query.
-        Returns raw objects returned from elastic-search.
-        """
+    def list(self, request):
+        # Raise 500 if elasticsearch is not running, or return an empty list ?
```
sarahfirstvoices, line 23: Is this still a question? I see the custom 500 error

firstvoices/firstvoices/settings.py
```diff
@@ -115,6 +119,18 @@
...
     INTERNAL_IPS = ["127.0.0.1", "10.0.2.2"]
+    LOGGING = {
+        "version": 1,
+        "disable_existing_loggers": False,
+        "handlers": {"console": {"class": "logging.StreamHandler"}},
+        "root": {"handlers": ["console"], "level": "WARNING"},
+        "loggers": {
+            ELASTICSEARCH_LOGGER: {
+                "handlers": ["console"],
+                "level": "INFO",  # Change level to INFO to view connection requests
```
sarahfirstvoices, line 130: Is this the level we want in production, or does this belong in settings.debug?

sarahfirstvoices (approved):

PR #138, opened 2023-05-26
FW-4125 Recalculate Custom Sort Task v2

bronson-c:
### Description of Changes
This PR adds an endpoint to trigger a custom order recalculation for a site's dictionary entries.

Sending a POST request to api/1.0/sites/<site-slug>/dictionary-cleanup/ will start a task to recalculate the custom order of the whole dictionary which runs asynchronously. The results of this recalculation are returned by the POST request.

Sending a GET request to the same url will return the status of the most recent recalculation task.

### Checklist

### Additional Notes
The results of the POST request are in the following format:
```
"recalculationResults": [
        {
            "title": "test",
            "cleanedTitle": "test",
            "previousCustomOrder": "⚑t⚑e⚑s⚑t",
            "newCustomOrder": "!$%!"
        },
        {
            "title": "ᐱᐱᐱ",
            "cleanedTitle": "AAA",
            "previousCustomOrder": "⚑ᐱ⚑ᐱ⚑ᐱ",
            "newCustomOrder": "###"
        }
    ]
```

and the results of the GET request are in this format:
```
"currentRecalculationTaskStatus": "SUCCESS"
```


firstvoices/backend/rules.py
```diff
@@ -0,0 +1,5 @@
+import rules
+
+from backend.permissions.predicates.base import is_superadmin
+
+rules.add_perm("views.has_custom_order_access", is_superadmin)
```
bronson-c, line 5: As the recalculate action is not tied to a model, I have opted for re-adding the superadmin rule. Additionally the recalculate view works the way the old recalculate preview view did before it was re-factored into a model view set. I am open to re-factoring the recalculate view as well if the team would prefer.

firstvoices/backend/tasks/alphabet_tasks.py
```diff
@@ -48,3 +48,28 @@ def recalculate_custom_order_preview(site_slug: str):
...
+        entry.save()
+        result["new_custom_order"] = entry.custom_order
+        result["cleaned_title"] = entry.title
+        if (
+            result["previous_custom_order"] != result["new_custom_order"]
+            or result["cleaned_title"] != result["title"]
+        ):
+            results.append(result)
+
+    return results
```
bronson-c, line 75: The recalculate action returns the results of the action in the following format:
```
"recalculationResults": [
        {
            "title": "test",
            "cleanedTitle": "test",
            "previousCustomOrder": "⚑t⚑e⚑s⚑t",
            "newCustomOrder": "!$%!"
        },
        {
            "title": "ᐱᐱᐱ",
            "cleanedTitle": "AAA",
            "previousCustomOrder": "⚑ᐱ⚑ᐱ⚑ᐱ",
            "newCustomOrder": "###"
        }
    ]
```
@caforbes is this a suitable response or would you prefer it look identical to the preview action's results?

firstvoices/backend/views/custom_order_recalculate_views.py
```diff
@@ -54,3 +63,55 @@ def create(self, request, *args, **kwargs):
 
         except recalculate_custom_order_preview.OperationalError:
             raise CeleryError()
+
+
+class CustomOrderRecalculateView(APIView):
```
bronson-c, line 68: As mentioned this view works the same as the first iteration of the preview view. I am open to refactoring it if the team would like this to be more consistent with our other APIs. However, I'd need a bit of advice for how to do so as there is no model associated with this API.

firstvoices/backend/views/custom_order_recalculate_views.py
```diff
@@ -54,3 +63,55 @@ def create(self, request, *args, **kwargs):
...
+
+        # Call the recalculation task
+        try:
+            recalculation_results = recalculate_custom_order.apply_async((site_slug,))
+            task = {"task_id": recalculation_results.task_id, "site_slug": site_slug}
+            self.tasks.append(task)
+
+            return Response(
+                {"recalculation_results": recalculation_results.get(timeout=360)},
+                status=200,
```
sarahfirstvoices, line 113: Wait, how does this return the final result if the task is running asynchronously?

firstvoices/backend/urls.py
```diff
@@ -38,7 +40,12 @@
 
 app_name = "api"
 
-urlpatterns = []
+urlpatterns = [
+    path(
+        "sites/<str:site_slug>/dictionary-cleanup/",
+        CustomOrderRecalculateView.as_view(),
+    ),
+]
```
sarahfirstvoices, line 48: This should be added to the sites_router like the other site content urls

firstvoices/backend/views/custom_order_recalculate_views.py
NolanVH: I have the same question. Unless I am missing something it looks like if the recalculation takes longer than the timeout it will raise an error?

firstvoices/backend/views/base_views.py
bronson-c, line 1: This file is originally from the category create API PR: https://github.com/First-Peoples-Cultural-Council/fv-be/pull/152. The new `FVPermissionViewSetMixin` was used to create the `CustomOrderFVPermissionViewSetMixin`. Thanks @sarahfirstvoices this really helped figure out the non member user permissions issue.

firstvoices/backend/views/base_views.py
```diff
@@ -1,25 +1,113 @@
...
+            return
+
+        # Determine whether we've to check object permissions (for detail actions)
+        obj = None
+        extra_actions = self.get_extra_actions()
+        # We have to access the unbound function via __func__
+        if handler.__func__ in extra_actions:
+            if handler.detail:
+                obj = self.get_object()
+            elif "create" in self.action:
```
bronson-c, line 95: I made this change so that any additional action contain the word "create" would be able to use the `get_object_for_create_permission` function. In my case it would be fore "create_preview" but it could apply to other additional actions in the future.

firstvoices/backend/views/custom_order_recalculate_views.py
```diff
@@ -1,51 +1,154 @@
...
+            return self.get_paginated_response(serializer.data)
+
+        # non-paginated response
+        serializer = self.serializer_class(
+            queryset, many=True, context={"request": request}
+        )
+        return Response(serializer.data)
+
+    @get_preview.mapping.post
+    @action(detail=False, methods=["post"])
```
bronson-c, line 130: This decorator creates a duplicate endpoint `api/1.0/sites/test-site/dictionary-cleanup/create_preview/` which only takes a POST request and acts the same as `dictionary-cleanup/preview/`. The decorator is required for the method to function as a non-detail endpoint. If we wanted to get rid of the duplicate endpoint I could potentially combine `get_preview` and `create_preview` into the same method with different actions depending on the request type. I'd like opinions on whether or not to go in this direction

firstvoices/backend/tasks/alphabet_tasks.py
```diff
@@ -48,3 +53,58 @@ def recalculate_custom_order_preview(site_slug: str):
...
+
+        if result["previous_custom_order"] != new_order:
+            result["new_custom_order"] = new_order
+        if result["title"] != cleaned_title:
+            result["cleaned_title"] = cleaned_title
+            result["is_title_updated"] = True
+
+        # Count unknown characters remaining in each entry, first split by character, then apply custom order
+        if "⚑" in new_order:
+            chars = alphabet.get_character_list(entry.title)
```
sarahfirstvoices, line 96: You could pull lines 96-98 into a little function to get the split custom order chars for a word, right? Might help with that code smell, and it would be easier to read.

firstvoices/backend/tasks/alphabet_tasks.py
```diff
@@ -48,3 +53,58 @@ def recalculate_custom_order_preview(site_slug: str):
...
+            for char in chars:
+                custom_order = alphabet.get_custom_order(char)
+                if "⚑" in custom_order:
+                    if custom_order not in unknown_character_count:
+                        unknown_character_count[custom_order] = 0
+                    unknown_character_count[custom_order] += 1
```
sarahfirstvoices, line 102: You could also pull lines 100-102 into a little function like add_or_increment_item(dictionary, key)

firstvoices/backend/tasks/alphabet_tasks.py
```diff
@@ -48,3 +53,58 @@ def recalculate_custom_order_preview(site_slug: str):
...
+
+        if result["new_custom_order"] or result["cleaned_title"]:
+            updated_entries.append(result)
```
sarahfirstvoices, line 105: If you moved this conditional up before the unknown character checks, I think you could pull the entirety of lines 74 to about 93 into a function that returns either a result or None, and then if it exists, append it to updated_entries.

sarahfirstvoices: Or even just put those lines in a function that always returns a result, and keep these lines in this function.

firstvoices/backend/tests/test_apis/test_dictionary_cleanup_api.py
```diff
@@ -2,64 +2,94 @@
...
-        )
+    def get_cleanup_endpoint(self, key, is_preview):
+        if is_preview:
+            return self.get_detail_endpoint(key) + "dictionary-cleanup/preview/"
+        else:
+            return self.get_detail_endpoint(key) + "dictionary-cleanup/"
 
     @pytest.mark.django_db
-    def test_recalculate_preview_get_404(self):
+    @pytest.mark.parametrize("is_preview", [True, False])
```
sarahfirstvoices, line 27: Rather than having all these tests check two different endpoints based on a flag, let's have two different subclasses with different settings and shared logic. Same amount of reuse, but it's much easier to understand and reason about when each class has one job instead of two

firstvoices/backend/views/base_views.py
sarahfirstvoices: Oh, gotcha. Good point about custom permissions. This is fine for now, but I am pondering how we might handle other permission names in the future.

firstvoices/backend/views/custom_order_recalculate_views.py
```diff
@@ -1,51 +1,154 @@
...
+
+            # Save the result to the database
+            CustomOrderRecalculationResult.objects.create(
+                site=site[0],
+                latest_recalculation_result=recalculation_results.get(timeout=360),
+                task_id=recalculation_results.task_id,
+                is_preview=False,
+            )
+
+            return Response({"message": "Recalculation has been queued."}, status=201)
```
sarahfirstvoices, line 100: Turns out there is actually a status code for "accepted/queued", so we could use 202 here.

NolanVH (commented):
This is looking great! Once the other comments on this PR are addressed I will re-review as needed. I have tested the cleanup locally and everything seems to be working well.

One minor thing I noticed is that the preview and non-preview views for dictionary-cleanup share the same `latestRecalculationDate` field name, which might cause some confusion. It might be worth changing the preview field to something like `latestRecalculationPreviewDate`.

PR #159, opened 2023-06-02
FW-4417 data migration import script

caforbes:
### Description of Changes
Added a script `scripts/import_csv_data.py` to perform imports from exported Nuxeo data. The script downloads the data files from AWS, drops all existing Sites from the db, and imports the new data before finally deleting the files (if successful). Status of the import is tracked in a new model `AppImportStatus` and specific warnings are logged along the way (this could be converted to async results later or left as is). For now, the import halts ungracefully if it encounters an error.

Adds a new dependency, `django-import-export`, to handle the file imports. This allows us to generally import columns directly to model fields, with custom tweaking provided in a new "resource" class. I added `SiteResource` for some custom tweaks (e.g. map users by their email field).

### Checklist

### Additional Notes
Not sure if I should update the readme since this is a special use script.


firstvoices/backend/models/app.py
```diff
@@ -56,3 +56,24 @@ class Meta:
...
         return f"{self.user} ({self.get_role_display()})"
+
+
+class AppImportStatus(BaseModel):
+    """Represents the status of manual data imports, such as nuxeo imports."""
+
+    label = models.CharField(max_length=150)
+    successful = models.BooleanField(default=False)
+    warnings = models.BooleanField(default=False)
+
```
NolanVH, line 67: Optional: It might be useful to add one more field which could hold some warning or failure information. This could probably be done down the road when we add to the import script though.

firstvoices/backend/admin/app_admin.py
```diff
@@ -35,3 +35,18 @@ class AppMembershipAdmin(BaseAdmin):
...
     )
+
+
+@admin.register(AppImportStatus)
+class AppImportStatusAdmin(BaseAdmin):
+    fields = (
+        "label",
+        "warnings",
+        "successful",
+    )
```
NolanVH, line 46: Adding these three fields as [readonly_fields](https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.readonly_fields) might be good here to make sure we don't accidentally change these values in the admin UI.

firstvoices/backend/resources/utils/import_export_widgets.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def __init__(self, create=False, *args, **kwargs):
+        self.create = create
+        super().__init__(model=User, field="email", *args, **kwargs)
+
+    def clean(self, value, row=None, **kwargs):
+        if self.create:
+            user, _ = self.model.objects.get_or_create(
+                **{self.field: value}, defaults={"id": value}
+            )
```
NolanVH, line 41: It looks like this is creating users without a password, meaning the password will be NULL in the database. I think this is fine for now and we can update this in the future when we have a way for users to claim migrated accounts, but we should probably make a note of this in one of the user management epic tickets. 

Additionally, the join date on the user model is getting set to the import time and we may want to consider migrating a join date if that exists in the data.

firstvoices/scripts/import_csv_data.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # raise errors to halt the import if an issue occurs
+        result = resource.import_data(dataset=table, dry_run=False, raise_errors=True)
+        logger.info(
+            " ".join([f"{type}: {total}" for type, total in result.totals.items()])
+        )
+
+
+for file in unmatched_files:
+    logger.warn(f"\n{file} not imported (no resource defined)")
```
NolanVH, line 86: I love the logs throughout the import script. This will make it much easier to trace how far an import made it before a failure.

NolanVH (approved):
The import script is looking great! I have tested it locally and everything is working as intended. This will be a good starting point to add other data migrations down the road.

Aside from the few minor comments I left, these changes all look good to me! Will approve now as I don't think the suggestions need re-review. Thanks @caforbes!

firstvoices/backend/resources/utils/import_export_widgets.py
caforbes: Definitely agreed. I will create-or-update a jira ticket about migration of user accounts and metadata to django specifically. We'll also need to think about collisions and fixes for the non-case-sensitive emails, which I didn't handle here.

PR #166, opened 2023-06-05
Fw 4379 add site parameter to search api

sahil97:
### Description of Changes
- Added `site` as a url parameter to the search API.
- returning site if site_slug is provide in serializers/utils.py
- Added site_slug to context when hydrating dictionary objects
- Added a file called query_builder_utils
- Returning Q objects from sub-query functions and assembling them at top level
- Added a baseViewSet for search which has a get_search_params functions which defaults to empty value for site, and later will return default for other params but should be overridden in specific classes such as SiteViewSet which adds the site param to this function's return value

### Checklist

### Additional Notes
N/A


firstvoices/backend/serializers/fields.py
```diff
@@ -27,7 +27,10 @@ def get_url(self, obj, view_name, request, format):
         kwargs = {self.lookup_url_kwarg: lookup_value}
 
         # add the site lookup
-        kwargs.update({"site_slug": get_site_from_context(self).slug})
+        if get_site_from_context(self):
```
sahil97, line 30: check If the site_slug is not present in the url, such as in the case of search API

firstvoices/backend/search/query_builder.py
```diff
@@ -32,10 +41,14 @@ def get_cleaned_search_term(q):
...
 
-def get_search_query(q=""):
+def get_search_query(q="", site_slug=""):
     search_query = get_search_object()
     search_term = get_cleaned_search_term(q)
 
+    if site_slug:
+        site_slug = validate_site(site_slug)
+        search_query = search_query.filter("term", site_slug=site_slug)
+
```
sarahfirstvoices, line 51: This seems like a good time to split out the different parts of the query into separate functions. Like get_site_query and get_term_query could return Q objects, and the job of get_search_query could be to assemble them based on the various conditions and inputs

firstvoices/backend/serializers/fields.py
```diff
@@ -27,7 +27,11 @@ def get_url(self, obj, view_name, request, format):
         kwargs = {self.lookup_url_kwarg: lookup_value}
 
         # add the site lookup
-        kwargs.update({"site_slug": get_site_from_context(self).slug})
+        try:
+            kwargs.update({"site_slug": get_site_from_context(self).slug})
+        except AttributeError:
+            # If site not present
+            kwargs.update({"site_slug": None})
```
sarahfirstvoices, line 34: Why is this necessary? This field is only meant to be used for site content urls that always have a site slug

firstvoices/backend/serializers/utils.py
```diff
@@ -1,2 +1,5 @@
 def get_site_from_context(serializer):
-    return serializer.context["site"]
```
sarahfirstvoices, line 2: Why would this ever be called without a site in context?

firstvoices/backend/views/search/search_views.py
```diff
@@ -56,14 +63,17 @@ def get_search_params(self):
         Function to return search params in a structured format.
         """
         input_q = self.request.GET.get("q", "")
+        site_slug = self.request.GET.get("site_slug", "")
 
-        return {"q": input_q}
+        return {"q": input_q, "site_slug": site_slug}
```
sarahfirstvoices, line 68: For the url parameter, let's just call this "site"

firstvoices/backend/serializers/fields.py
sahil97: Yes, So It seems that the error is arising due to the usage of the `DictionaryDetailSerializer`. This serializer tries to add the `url` field, which requires the site to be in context in order to form the URL. However, in the case of the search API, which is a top-level API, the site is not included in the URL but passed as a parameter. As a result, an exception is thrown.
To address this problem, I implemented a workaround in the `utils.py `file. The conditional clause there checks if the site is present in the context and returns None if it is not. However, accessing this `None` value may result in an AttributeError, So I also added a try/catch block to handle that, as per Sonar's suggestion.

sarahfirstvoices: If this is called with no site_slug available, it will fail to correctly reverse the url though, right? So it seems strange to prevent the error when it is an actual problem

sarahfirstvoices: I think we need to solve this a different way, and intentionally make the DictionaryEntrySerializer work outside of a site content api.

firstvoices/backend/search/query_builder.py
sahil97: I've refactored the main function into smaller functions, but the logic to assemble Q objects still rest inside the specific functions. The parent function checks if the param is present, if yes, it passes the param to the required function and expects a modified search query in return with added filter/query.

firstvoices/backend/search/utils/query_builder_utils.py
```diff
@@ -0,0 +1,80 @@
...
+        ],
+        minimum_should_match=1,
+    )
+
+
+def get_site_filter_query(site_slug):
+    # Validate site exists
+    site_count = Site.objects.filter(slug=site_slug).count()
+    if site_count == 0:
+        raise Http404
```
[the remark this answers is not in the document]
sahil97: Yes, It is also handled in the view as I'm using the SiteContentViewSetMixin which has a method to get_validated_site which I'm using to pass the site further. I added this as an extra validation step but can be removed.

firstvoices/backend/serializers/utils.py
```diff
@@ -1,5 +1,13 @@
...
+
 def get_site_from_context(serializer):
     if "site" in serializer.context:
         return serializer.context["site"]
-    else:
-        return None
+    elif "site_slug" in serializer.context:
+        site = Site.objects.filter(slug=serializer.context["site_slug"])
+        if len(site) == 0:
+            return Http404
```
sarahfirstvoices, line 12: It's great to check whether the site was found here, but I don't think a 404 is the right error. This is way down in a serializer util, we don't have enough context to know what should be returned to the user. This situation is most likely due to a programming error here. I think we should log an error that says something like "Tried to retrieve a site from context but no site found for the site_slug provided.", and then raise an APIException

sarahfirstvoices: We should also log and raise if we get past the two conditions and haven't returned anything

firstvoices/backend/urls.py
```diff
@@ -35,6 +38,7 @@
     CustomOrderRecalculatePreviewView,
     basename="dictionary-cleanup/preview",
 )
+sites_router.register(r"search", SiteSearchViewsSet, basename="site-search")
```
sarahfirstvoices, line 41: yay!

firstvoices/backend/views/search/base_search_views.py
```diff
@@ -63,11 +56,11 @@ def get_search_params(self):
         Function to return search params in a structured format.
         """
         input_q = self.request.GET.get("q", "")
-        site_slug = self.request.GET.get("site_slug", "")
 
-        return {"q": input_q, "site_slug": site_slug}
+        search_params = {"q": input_q, "site_slug": ""}
```
sarahfirstvoices, line 60: Minor: we could just leave out the site slug here instead of setting it to blank

firstvoices/backend/search/query_builder.py
```diff
@@ -1,100 +1,34 @@
...
-
-def get_cleaned_search_term(q):
-    """
-    clean incoming string.
-    case-sensitivity handled by analyzer in the search document.
-    """
-    return q.strip()
-
-
 def get_search_query(q="", site_slug=""):
```
sarahfirstvoices, line 19: Any reason not to default these to None?

sarahfirstvoices (approved):
Just those little error-handling changes and then I think this is good to go!

PR #175, opened 2023-06-07
FW-4422 add doc types parameter to search api

sahil97:
### Description of Changes
Added a types parameter to search API. accepts a comma-separated string as input. Ignores invalid types and return results from all types if empty input is passed.

### Checklist
- [x] README / documentation has been updated if needed~

### Additional Notes
N/A


bronson-c (commented):
Looking good! Just a minor change I found and a question about tests: 

Would it be possible to add search API tests to check the full process? For example creating the dictionary entries for a site, then searching for those entries with filter parameters and ensuring the correct doctypes are being filtered out.

firstvoices/backend/search/query_builder.py
```diff
@@ -9,17 +11,17 @@
 )
 
 
-def get_search_object():
+def get_search_object(indices):
     client = Elasticsearch()
```
sarahfirstvoices, line 15: Needs a merge with Rob's recent change

firstvoices/backend/search/query_builder.py
```diff
@@ -33,4 +35,14 @@ def get_search_query(q=None, site_slug=None):
     if site_slug:
         search_query = search_query.query(get_site_filter_query(site_slug))
 
+    # Adding type filters
+    # If only one of the "words" or "phrases" is present, we need to filter out the other one
+    # no action required if both are present
```
sarahfirstvoices, line 40: Minor, but I think all this conditional logic could go in a get_types_query function.

firstvoices/backend/search/query_builder.py
```diff
@@ -9,17 +11,17 @@
...
-def get_search_object():
+def get_search_object(indices):
     client = Elasticsearch()
-    indices = get_indices()
     s = Search(using=client, index=indices)
     return s
 
 
-def get_search_query(q=None, site_slug=None):
+def get_search_query(q=None, site_slug=None, types=VALID_DOCUMENT_TYPES):
```
sarahfirstvoices, line 20: This is a great way to do the default value for types

firstvoices/backend/search/utils/query_builder_utils.py
```diff
@@ -3,17 +3,42 @@
...
+    selected_values = [
+        value.strip().lower()
+        for value in values
+        if value.strip().lower() in allowed_values
+    ]
+
+    if len(selected_values) == 0:
+        return None
+
+    return selected_values
```
sarahfirstvoices, line 25: Love this. A very clear, concise way to do this fussy set of conditions

firstvoices/backend/search/utils/query_builder_utils.py
```diff
@@ -3,17 +3,42 @@
...
     SONG = ELASTICSEARCH_SONG_INDEX
     STORY = ELASTICSEARCH_STORY_INDEX
     """
-    list_of_indices = [ELASTICSEARCH_DICTIONARY_ENTRY_INDEX]
-    return list_of_indices
+    indices = set()
+
+    for doc_type in types:
+        if doc_type == "WORD" or doc_type == "PHRASE":
+            indices.add(ELASTICSEARCH_DICTIONARY_ENTRY_INDEX)
```
sarahfirstvoices, line 39: Extremely minor: "doc_types" are from nuxeo. These can just be types

firstvoices/backend/search/query_builder.py
```diff
@@ -33,4 +35,14 @@ def get_search_query(q=None, site_slug=None):
...
 
+    # Adding type filters
+    # If only one of the "words" or "phrases" is present, we need to filter out the other one
+    # no action required if both are present
+    if "words" in types and "phrases" not in types:
+        search_query = search_query.query(
+            ~Q("match", type=TypeOfDictionaryEntry.PHRASE)
+        )
+    elif "phrases" in types and "words" not in types:
+        search_query = search_query.query(~Q("match", type=TypeOfDictionaryEntry.WORD))
```
sarahfirstvoices, line 46: Is there a reason to do the filtering as "is not a phrase" instead of "is a word"?

firstvoices/backend/views/search/base_search_views.py
```diff
@@ -38,12 +39,49 @@
...
+                    ),
+                    OpenApiExample(
+                        "words",
+                        value="words, phrases",
+                        description="specifically looks for documents in the words document type.",
+                    ),
+                    OpenApiExample(
+                        "words, invalid_type",
+                        value="words",
+                        description="filters out the invalid document types and returns results only "
```
sarahfirstvoices, line 74: Rather than saying this "filters out", let's say it "ignores"

firstvoices/backend/views/search/base_search_views.py
```diff
@@ -57,15 +95,24 @@ def get_search_params(self):
...
+
+        search_params = {"q": input_q, "types": valid_types_list, "site_slug": ""}
         return search_params
 
     def list(self, request, **kwargs):
         search_params = self.get_search_params()
 
+        # If no valid types are passed, return emtpy list as a response
+        if not search_params["types"]:
+            return Response(data=[])
```
sarahfirstvoices, line 109: Is this correct? If they don't specify types, I would expect all data types to be returned.

sarahfirstvoices (approved):

NolanVH (approved):
Aside from the issues addressed in the comments by others, these changes all look good to me. Thanks @sahil97.

sahil97:
> Looking good! Just a minor change I found and a question about tests:
> 
> Would it be possible to add search API tests to check the full process? For example creating the dictionary entries for a site, then searching for those entries with filter parameters and ensuring the correct doctypes are being filtered out.

Yes, adding those tests are certainly possible, but running those tests require us to add a elasticsearch server either to github CI or provide a external ES server which can be queried when the tests are running. It was discussed earlier that since ES library is well tested, we should test the queries being sent which is our current plan.

firstvoices/backend/search/utils/query_builder_utils.py
sahil97: Yes, that was my first attempt as well haha. So the "type" variables gives a warning prompt of it shadowing a internal variable. To avoid that, I went with this name. But I can always rename it if required. This name is only used here and then to test this functionality.

firstvoices/backend/search/query_builder.py
sahil97: Yes, so adding in "is a word" instead of "is not a phrase" will result in more modification of queries while this will require minimal modification even when we add songs and stories due to each document type having their own index

firstvoices/backend/views/search/base_search_views.py
sahil97: so due to the default behaviour of get_search_params function, if a user does not specify types, it will add all types to search_params["types"].
The above case will only be triggered, when the user has specified all invalid types, so the get_search_params function makes `search_params["types"]=None` which will return empty results.

PR #176, opened 2023-06-07
FW-4420 add search pagination

bronson-c:
### Description of Changes

- Adds dynamic pagination by slicing the ES query according the the expected values on each page.
- Bonus change: adds custom order sorting to search.

### Checklist

### Additional Notes
N/A


firstvoices/backend/views/search/base_search_views.py
```diff
@@ -79,4 +87,11 @@ def list(self, request, **kwargs):
         # Adding data to objects
         hydrated_objects = hydrate_objects(search_results, request)
 
+        page = self.paginate_queryset(hydrated_objects)
+        # TODO: Use dictionary entry serializer to properly serialize search results
```
bronson-c, line 91: Currently pagination is returning all hydrated objects on multiple pages, rather than splitting them into the multiples pages as expected. Once the hydrated objects are properly serialized (needs camelcase to underscore conversion) pagination should function as expected)

firstvoices/backend/views/search/base_search_views.py
```diff
@@ -70,7 +81,7 @@ def list(self, request, **kwargs):
 
         # Get search results
         try:
-            response = search_query.execute()
+            response = search_query[0:ES_MAX_RESULTS].execute()
```
sarahfirstvoices, line 84: This looks like we will always be retrieving the first 100 search results. Is that right? I would expect that for page 2 with the default page size, we'd be retrieving more like [26-50] here

firstvoices/backend/views/search/base_search_views.py
```diff
@@ -79,4 +90,8 @@ def list(self, request, **kwargs):
         # Adding data to objects
         hydrated_objects = hydrate_objects(search_results, request)
 
+        page = self.paginate_queryset(hydrated_objects)
+        if page is not None:
+            return self.get_paginated_response(data=page)
+
```
sarahfirstvoices, line 96: Is this going to scale if someone has 30 pages of search results? I think we'll need to retrieve the page of results directly from elastic search, rather than retrieving extra results and then getting django to paginate them for us

firstvoices/backend/search/utils/constants.py
```diff
@@ -10,6 +10,12 @@
     "operation: %s. %s id: %s"
 )
 
+# Page size
+ES_PAGE_SIZE = 25
+
+# Number of results to return
+ES_MAX_RESULTS = 100
```
sarahfirstvoices, line 17: To clarify, 100 is the maximum page size from the old server, not a cap on the total search results we can return.

firstvoices/backend/views/search/base_search_views.py
sahil97: I believe we should refrain from altering the length of the list of results outside Elasticsearch, as it may potentially lead to inconsistent outcomes.
For instance, if Elasticsearch returns 25 results for the first page and the user requests only 20 results, omitting this parameter in the ES query could cause Django pagination to remove the last 5 results. Consequently, these 5 results would not be included on the next page either, as they do not exist on Elasticsearch's second page.
One way to avoid this would be to extract the pagination parameters from the request and pass them to the ES query to obtain paginated results directly from Elasticsearch.

firstvoices/backend/search/indices/dictionary_entry_document.py
```diff
@@ -26,6 +26,8 @@ class DictionaryEntryDocument(Document):
     document_id = Text()
     site_id = Keyword()
     full_text_search_field = Text()
+
+    custom_order = Keyword()
```
sahil97, line 30: minor: Since this is dictionary entry specific field, can be moved a few lines down

firstvoices/backend/views/search/base_search_views.py
```diff
@@ -206,8 +208,28 @@ def get_search_params(self):
...
+        page = int(self.request.GET.get("page", 1))
+        page_size = int(self.request.GET.get("pageSize", default_page_size))
+        start = (page - 1) * page_size
+        end = start + page_size
+
+        pagination_params = {
+            "page_size": page_size,
+            "page": page,
+            "start": start,
+            "end": end,
```
sahil97, line 226: minor: pagination_params["end"] can be removed since we are using from and size in the ES query.

sahil97 (approved):
The custom_order field should also be added in the management/commands/_helper.py - dictionary_entry_iterator() function else that field may be missed when index is being rebuilt. 
Aside from the above change and a few minor comments, PR looks good to go. Nice work!

PR #197, opened 2023-06-14
Fw 4434 add management command to rebuild the index with current data

sahil97:
### Description of Changes

### Checklist

### Additional Notes
To test locally:
1. Ensure that elastic-search server is running locally. Instructions can be found in the readme to get the docker container running.
2. Run the following command from firstvoices folder:
`python3 manage.py rebuild_index`
3. It should delete the existing index, and recreate the index with all the dictionary entry documents present in the database.
4. Verify that the documents are indexed by searching for them. All documents can be fetched if `search_term` is not provided, or you can try randomly search for some entries.


firstvoices/backend/management/commands/_helper.py
```diff
@@ -0,0 +1,56 @@
...
+        print("Current index not found for deletion. Creating a new index.")
+
+
+def dictionary_entry_iterator():
+    queryset = DictionaryEntry.objects.all()
+    for entry in queryset:
+        (
+            translations_text,
+            part_of_speech_text,
+        ) = get_translation_and_part_of_speech_text(entry)
```
sahil97, line 45: It took less than a 4 seconds for <10 documents to be indexed, since the deletion, creation takes a few seconds as well. The bulk API to index should handle large queryset size, and since iterator is being used, memory should not be a problem as well.
The only place where I feel this can be optimized further is these 2 lines where we are fetching translation and notes after fetching the object. 
Leaving this note here for reference.

firstvoices/backend/management/commands/_helper.py
```diff
@@ -0,0 +1,112 @@
...
+                "rebuilding."
+            )
+    except Exception as e:
+        # Any other exception due to which we are not able to build a new index or add documents to it
+        # Attach alias back to current index so search keeps working
+        add_alias(current_index, index_name)
+        raise CommandError(
+            "The following error occurred while indexing documents. Adding alias back to current index. "
+            + str(e)
+        )
```
sarahfirstvoices, line 59: I'm still a bit confused about the handling of the aliases here. Do we need to delete the alias for the current index before trying to build the new one? It seems like we will still have some search downtime using this method, because the alias is going to point to an incomplete index while it is building. (On our old server, the rebuild can take over an hour.)

sarahfirstvoices (approved):

PR #219, opened 2023-06-20
FW-4395 Fix API docs warnings

NolanVH:
### Description of Changes
These changes add missing parameters and field types to the `@extend_shema_...` decorators to remove the majority of the warnings when loading the API docs page. Note that nothing has visibly changed as [drf-spectacular](https://github.com/tfranzel/drf-spectacular) will automatically fall back on defaults, but the warnings are now gone. In the future, the API docs page could be updated with things like detailed examples to closely match the expected responses.

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/serializers/site_serializers.py
+++ b/firstvoices/backend/serializers/site_serializers.py
@@ -1,3 +1,5 @@
+from drf_spectacular.types import OpenApiTypes
+from drf_spectacular.utils import extend_schema_field, extend_schema_serializer
 from rest_framework import serializers
 
 from backend.models.app import AppJson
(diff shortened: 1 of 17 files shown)


firstvoices/backend/serializers/site_serializers.py
```diff
@@ -42,6 +44,22 @@ class Meta(LinkedSiteSerializer.Meta):
         fields = LinkedSiteSerializer.Meta.fields + ("logo", "features")
 
 
+@extend_schema_serializer(
```
sahil97, line 47: would this exclude these fields from the docs page ? can we not specify types for each fields as well ?
like list of category model in categories

firstvoices/backend/views/custom_order_recalculate_views.py
```diff
@@ -26,6 +32,11 @@
             403: OpenApiResponse(description="Todo: Not authorized for this Site"),
             404: OpenApiResponse(description="Todo: Site not found"),
         },
+        parameters=[
+            OpenApiParameter(
```
sahil97, line 36: should we add this OpenAPIParameter to a variable ? will help if we need to modify the param later since its being reused throughout the file

firstvoices/backend/serializers/site_serializers.py
NolanVH: Yes, it does exclude these fields from the docs page. Currently, they are not being generated anyway and a warning was being thrown. I agree that we should go through the docs and fill them in completely, but I think this should be a low priority right now. One option would be to close this PR and leave the mass of warnings as a reminder to update the docs later.

sahil97 (commented):
LGTM!

sahil97 (approved):

firstvoices/backend/serializers/site_serializers.py
sarahfirstvoices: I think it's great to get rid of the warnings for now. @NolanVH could you please log a ticket in the backlog to deal with these missing fields in the future?

sarahfirstvoices (approved):

PR #235, opened 2023-06-27
FW-4440, FW-4563 import fixes: date/time metadata, less destructive

caforbes:
### Description of Changes

Created/last modified metadata now import correctly, using the dates/times from the CSV instead of autosaving new times on import. This breaks last-modified times for actual data modification but can be fixed post-prod-migration and pre-launch.

And, now in advance of importing sites, any sites in the BE with the same UID will be deleted (and no other sites) - less destructive.

I also added a setting so importing permissions are tied to update permissions, but can roll this back if it is a blocker.

### Checklist

### Additional Notes
N/A


NolanVH (approved):
These changes all look good to me. Thanks @caforbes.

One thought I had is that it might be possible to keep a list of imported data IDs and dates together, import the data, and then update the date fields afterwards so that we don't have to get rid of `auto_now` and `auto_now_add`. It looks like the fields could be updated without using `Model.save()` as that is the only function that updates using `auto_now`. There is more information [in the Django docs](https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.DateField.auto_now). This might not be worth sinking a ton of time into right now though.

caforbes:
I see, so to import as normal at first with the auto_now values, and then immediately after do a bulk-style update from the same CSVs to overwrite the created/last_modified values without calling `save()`. I hadn't thought of doing it in that order. I think that could possibly work, but it would get potentially hairy with models where we are creating new objects on import (e.g. `DictionaryEntry` creates `Notes`, `Acknowledgements`). It would be possible but would require extra logic we weren't planning on.

Adding another fix here to import with a dummy user when an existing user isn't recognized in the data -- maybe @sarahfirstvoices could give a quick look and an opinion on this option?

firstvoices/backend/resources/utils/import_export_widgets.py
```diff
@@ -28,17 +27,31 @@ def render(self, value, obj=None):
...
             user, _ = self.model.objects.get_or_create(
                 **{self.field: value}, defaults={"id": value}
             )
             return user
         else:
-            return super().clean(value, row, **kwargs)
+            # for now, return a dummy user
+            dummy_user, _ = self.model.objects.get_or_create(
+                **{self.field: "test@test.com"}, defaults={"id": value}
+            )
```
sarahfirstvoices, line 56: Just for our own future reference, the ticket for really migrating the user data is FW-4446

sarahfirstvoices (approved):

sarahfirstvoices:
> so to import as normal at first with the auto_now values, and then immediately after do a bulk-style update...

I actually think the changes to the models are an improvement and we don't even need to change them back later. We still get the default time values, but we can override them if we have some reason to. Might come in handy again someday!

So the current way seems fine to me.

caforbes:
As I was writing a comment to say that unfortunately it doesn't work the way Sarah thinks it works, I realized it would be faster to make it so! 💡 So here we are, last-modified is no longer broken. You can provide manual values on creation and thereafter it will update the `last_modified` value on save. Thanks for the brainwave Sarah!

PR #244, opened 2023-06-29
FW-4328 Add widget models

NolanVH:
### Description of Changes
These changes add the following widget models:
- `Widget` (a base widget which can be used at the app level)
- `WidgetSettings` (settings key-value pairs with a foreign key to a widget)
- `SiteWidget` (inherits from `Widget` which has a site and visibility controls)
- `SiteWidgetList` (contains a many to many field relating to SiteWidget instances)
- `SiteWidgetListOrder` (The through model for the widgets many to many field on a `SiteWidgetList`. Contains an order field indicating the widget order within a `SiteWidgetList`.)

A `homepage` field has been added to the site model relating a single `SiteWidgetList` instance to a site to indicate the homepage widgets.

### Checklist

### Additional Notes
N/A


firstvoices/backend/models/sites.py
```diff
@@ -132,6 +133,14 @@ class Meta:
     # from fvdialect:contact_email
     contact_email = models.EmailField(null=True, blank=True)
 
+    homepage = models.OneToOneField(
+        SiteWidgetList,
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="site_homepage",
```
sarahfirstvoices, line 141: These reverse names are always hard. Maybe this should be something like "homepage_of" or "homepage_site" ?

firstvoices/backend/models/sites.py
```diff
@@ -132,6 +133,14 @@ class Meta:
     # from fvdialect:contact_email
     contact_email = models.EmailField(null=True, blank=True)
 
+    homepage = models.OneToOneField(
+        SiteWidgetList,
+        null=True,
+        blank=True,
```
sarahfirstvoices, line 139: I think for this one we don't need blank=True, since it isn't a string field

firstvoices/backend/models/widget.py
```diff
@@ -0,0 +1,149 @@
...
+from backend.models.base import (
+    BaseControlledSiteContentModel,
+    BaseModel,
+    BaseSiteContentModel,
+)
+from backend.permissions import predicates
+
+
+class WidgetTypes(models.IntegerChoices):
+    # enum intentionally has gaps to allow future changes to keep sequential order
```
sarahfirstvoices, line 14: Based on Guy's comment that this does not need to be constrained on the backend, I think the widget type could just be a string field. No need for an enum. (Maybe we'll add it someday if the backend starts to have structured widget data.)

firstvoices/backend/models/widget.py
```diff
@@ -0,0 +1,149 @@
...
+    WIDGET_TEXT = 100
+    WIDGET_TEXTCONCISE = 110
+    WIDGET_TEXTFULL = 120
+    WIDGET_TEXTICONS = 130
+    WIDGET_TEXTMULTI = 140
+    WIDGET_WOTD = 150
+
+
+class WidgetFormats(models.IntegerChoices):
+    # enum intentionally has gaps to allow future changes to keep sequential order
```
sarahfirstvoices, line 34: This is fine, but I'm not sure this one really needs gaps since there isn't really any meaning in sorting these descending vs ascending

firstvoices/backend/models/widget.py
```diff
@@ -0,0 +1,149 @@
...
+        verbose_name_plural = _("widgets")
+        rules_permissions = {
+            "view": rules.always_allow,
+            "add": predicates.is_superadmin,
+            "change": predicates.is_superadmin,
+            "delete": predicates.is_superadmin,
+        }
+
+    title = models.CharField(max_length=225)
+    type = models.IntegerField(
```
sarahfirstvoices, line 52: Is your IDE complaining that type is a built-in python function name as well? Probably worth changing this to widget_type or something

firstvoices/backend/models/widget.py
```diff
@@ -0,0 +1,149 @@
...
+            "view": predicates.has_visible_site,
+            "add": predicates.is_superadmin,  # permissions will change when we add a write API
+            "change": predicates.is_superadmin,
+            "delete": predicates.is_superadmin,
+        }
+
+    widgets = models.ManyToManyField(
+        SiteWidget, related_name="sitewidget_set", through="SiteWidgetListOrder"
+    )
+    title = models.CharField(max_length=225)
```
sarahfirstvoices, line 113: Does a widget list need a title?

firstvoices/backend/models/widget.py
```diff
@@ -0,0 +1,149 @@
...
+    class Meta:
+        verbose_name = _("site widget list order")
+        verbose_name_plural = _("site widget list orders")
+        constraints = [
+            models.UniqueConstraint(
+                fields=["order", "site_widget_list"], name="unique_widget_order"
+            ),
+        ]
+        rules_permissions = {
+            "view": rules.always_allow,
```
sarahfirstvoices, line 129: I think this should be has_visible_site

firstvoices/backend/models/sites.py
NolanVH: Good call. I will update this.

sarahfirstvoices (approved):
Looking good! No need to re-review after dealing with my minor comments

firstvoices/backend/models/widget.py
NolanVH: Good catch. Thanks!

firstvoices/backend/models/widget.py
NolanVH: Good catch, although it can be very hard to tell the lists apart without it, so it may be worth keeping this as an optional field?

bronson-c: Would the list's "site" fk not be enough to tell the lists apart? Or will sites have multiple widget lists?

NolanVH: A site can have multiple widget lists. The intention is for each page (when the page model is added) on a site to have a SiteWidgetList foreign key relation. A SiteWidgetList also specifies the homepage widgets for a site by a foreign key relation in the site model.

bronson-c (approved):
No major changes from me beyond what Sarah has already mentioned. Tested locally as well, looks great!

firstvoices/backend/models/widget.py
bronson-c: Ah I see, thanks for the clarification. Maybe we can nix the title field and when the page model is added we could use the page foreign key to differentiate the lists then?

NolanVH: That sounds good, thanks for the feedback.

firstvoices/backend/models/sites.py
NolanVH: Don't we need blank=True here to allow a site to be created with no homepage widgets set?

firstvoices/backend/models/widget.py
NolanVH: As discussed in a call we will leave this for now and investigate other options in FW-4368. An example of how to deal with permissions has been left in an example branch: https://github.com/First-Peoples-Cultural-Council/fv-be/blob/457043bc6902e3b0c8c0a37db4224db1f670891e/firstvoices/backend/models/widget.py#L76-L133

PR #246, opened 2023-06-30
FW-4418 Search Permissions

bronson-c:
### Description of Changes
- Adds view permission filters to the ES search query that function the same way as the `is_visible_object` filter in `backend/permissions/filters/view.py`
- Bonus change: adds `python manage.py rebuild_index` to the reset database script.

### Checklist

### Additional Notes
Apologies about the commit history, I had accidentally been working in the wrong branch for a bit.


firstvoices/backend/search/query_builder.py
```diff
@@ -20,6 +21,7 @@ def get_search_object(indices):
 
 
 def get_search_query(
+    user=None,
```
bronson-c, line 24: Possible security issue here? I had made user an optional paramater so that the other tests pass without major changes, but if an attacker were to spoof that there is no user then view permissions would not be active. However, I'm not sure if that's possible.

sarahfirstvoices: Better default would be AnonymousUser()

firstvoices/backend/permissions/filters/view.py
```diff
@@ -5,6 +5,8 @@
 #
 
 
+# Note that this view filter is re-created in the following file to work with ElasticSearch:
+# firstvoices/apps/fv_search/permissions.py
```
sarahfirstvoices, line 9: Love this but I think it can be even more intense. Can you please add some kind of direct instruction, like "If you change this, you need to make matching changes to the search permissions and to the predicate"

firstvoices/backend/search/utils/query_builder_utils.py
```diff
@@ -156,6 +159,55 @@ def get_site_filter_query(site_id):
     return Q("bool", filter=[Q("term", site_id=site_id)])
 
 
+def get_view_permissions_filter(user):
+    """
+    Re-creation of the is_visible_object filter from backend/permissions/filters/view.py
```
sarahfirstvoices, line 164: Please add an intense instruction here as well :)

firstvoices/backend/search/utils/query_builder_utils.py
```diff
@@ -156,6 +159,55 @@ def get_site_filter_query(site_id):
...
+    The logic is translated into an ES query
+    """
+
+    # base.is_at_least_staff_admin(user)
+    # if user has a staff app role or higher, they can see all objects
```
sarahfirstvoices, line 169: These comments are helpful-- nice!

sarahfirstvoices (approved):
Looking good! Just a few minor comments

firstvoices/backend/search/utils/query_builder_utils.py
```diff
@@ -156,6 +159,55 @@ def get_site_filter_query(site_id):
...
+
+    # base.has_public_access_to_obj(user)
+    filter_list.append(
+        Q("term", site_visibility=Visibility.PUBLIC)
+        & Q("term", visibility=Visibility.PUBLIC)
+    )
+
+    # add all the filters to the query
+    for f in filter_list:
+        query.should.append(f)
```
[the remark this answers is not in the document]
bronson-c: This is combining the filters using OR, so the documents that are viewable are (public docs) OR (membership docs from site x) OR (membership docs from site y) etc.

NolanVH (approved):
This is looking great! Tested working locally. Search is returning results filtered by memberships and permissions.

Aside from my minor comment about the reset script and the feedback from others, this is looking ready to me. Will approve now as there is no need to re-review after those minor changes have been made. Thanks @bronson-c.

firstvoices/reset-local-database.sh
```diff
@@ -141,6 +141,16 @@ case $yn in
...
+    # Rebuild Elasticsearch index
+    printf '\n'
+    printf 'Rebuilding Elasticsearch index.\n'
+    python $SCRIPT_DIR/manage.py rebuild_index
+    retval=$?
+    if [ $retval -ne 0 ]; then
+      printf "Elasticsearch index rebuild failed: exit code $retval\n"
+      exit $retval
+    fi
+
```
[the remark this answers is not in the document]
bronson-c: I've added a check for a local instance of ES before rebuilding the index. Nice catch!

sahil97:
Nice implementation. One small thing that I noticed missing is the update signal for sites, in a scenario where a site's visibility is changed, e.g. from 'team' to 'public', all the existing entries in the index needs to be updated with the new visibility else the search will produce incorrect results.
This will be a heavy operation since it requires fetching all entries from the index associated with a site. Similar issue was present with site_slug, and thus we moved to indexing site_id instead. Verification that visibility was changed if a site update signal is fired is recommended before continuing with the operation.

PR #269, opened 2023-07-06
FW-4213 Character Update API

bronson-c:
### Description of Changes

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/models/characters.py
+++ b/firstvoices/backend/models/characters.py
@@ -37,7 +37,7 @@ class Character(RelatedMediaMixin, BaseSiteContentModel):
         rules_permissions = {
             "view": predicates.has_visible_site,
             "add": predicates.is_superadmin,
-            "change": predicates.is_superadmin,
+            "change": predicates.is_at_least_assistant_or_super,
             "delete": predicates.is_superadmin,
         }
 
--- a/firstvoices/backend/serializers/character_serializers.py
+++ b/firstvoices/backend/serializers/character_serializers.py
@@ -1,9 +1,13 @@
 from rest_framework import serializers
+from rest_framework.relations import PrimaryKeyRelatedField
 
 from backend.models.characters import Character, CharacterVariant, IgnoredCharacter
+from backend.models.dictionary import DictionaryEntry
+from backend.serializers.base_serializers import UpdateSerializerMixin
 from backend.serializers.dictionary_serializers import DictionaryEntrySummarySerializer
 from backend.serializers.fields import SiteHyperlinkedIdentityField
 from backend.serializers.media_serializers import RelatedMediaSerializerMixin
+from backend.serializers.validators import SameSite, UniqueForSite
 
 
 class IgnoredCharacterSerializer(serializers.ModelSerializer):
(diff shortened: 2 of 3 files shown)


firstvoices/backend/serializers/dictionary_serializers.py
```diff
@@ -206,3 +206,34 @@ class Meta:
...
+
+
+class RelatedDictionaryEntrySerializerMixin(metaclass=serializers.SerializerMetaclass):
+    related_dictionary_entries = WritableRelatedDictionaryEntrySerializer(
+        required=False,
+        many=True,
+        queryset=dictionary.DictionaryEntry.objects.all(),
+    )
+
+    def validate(self, attrs):
```
bronson-c, line 228: I added this override method to be equivalent to the `SameSite` validator. During testing I found that specified validators are not applied in many to many nested fields. Verified in the documentation here: https://www.django-rest-framework.org/api-guide/validators/#updating-nested-serializers

sahil97 (approved):
Looks good. 
~~Does the `sameSite` validator with `PrimaryKeyRelatedField` causes some issues in case of `relatedDictionaryEntries` field ? ~~
Got it. Didn't read the comment before.

firstvoices/backend/models/characters.py
```diff
@@ -37,7 +37,7 @@ class Meta:
         rules_permissions = {
             "view": predicates.has_visible_site,
             "add": predicates.is_superadmin,
-            "change": predicates.is_superadmin,
+            "change": predicates.is_at_least_assistant_or_super,
```
caforbes, line 40: Everything to do with characters is restricted to only language admins, so this should probably be `is_language_admin_or_super`

firstvoices/backend/tests/test_apis/test_characters_api.py
```diff
@@ -142,3 +143,176 @@ def test_detail_related_entries(self):
...
+        site = factories.SiteFactory.create(visibility=Visibility.PUBLIC)
+        character = factories.CharacterFactory.create(site=site)
+        dictionary_entry = factories.DictionaryEntryFactory.create(site=site)
+
+        req_body = {
+            "note": self.CHARACTER_NOTE,
+            "related_dictionary_entries": [str(dictionary_entry.id)],
+        }
+
+        user = factories.get_app_admin(role=AppRole.SUPERADMIN)
```
caforbes, line 228: Perhaps it would be better to check that these tests work for the language admin role? (or both)

firstvoices/backend/tests/test_apis/test_characters_api.py
```diff
@@ -142,3 +143,176 @@ def test_detail_related_entries(self):
                 "type": entry1.type,
             }
         ]
+
+    @pytest.mark.django_db
+    def test_update_related_media(self):
```
caforbes, line 148: naive Q: would it be worthwhile to have a test where only 1 related media was updated, instead of all of them? (in light of the required=false change you made above) Or is this already covered by other tests?

caforbes (approved):
Suggested tweaks to permissions, which should be quick to fix. Otherwise looks good to me. :)

firstvoices/backend/tests/test_apis/test_characters_api.py
bronson-c: Good point! I'll split this test into 3

PR #290, opened 2023-07-12
FW-4560 Add SitePage model

NolanVH:
### Description of Changes
This PR adds the SitePage model.

### Checklist

### Additional Notes
A check constraint has been added to the model which ensures that an instance can have either no banner, or one of an image banner or video banner, but not both at the same time.


firstvoices/backend/models/page.py
```diff
@@ -0,0 +1,65 @@
...
+from backend.models.widget import SiteWidgetList
+from backend.permissions import predicates
+
+
+class SitePage(BaseControlledSiteContentModel):
+    class Meta:
+        verbose_name = _("site page")
+        verbose_name_plural = _("site pages")
+        rules_permissions = {
+            "view": predicates.is_visible_object,
```
NolanVH, line 18: I would love to hear what the team thinks about the permissions for a page here. Currently, in this PR a page has its own visibility controls. Should a page's visibility be tied to the site visibility (if the site is visible then the page is visible) or is it helpful to have separate access control like this? @gmcauliffe @sarahfirstvoices 

Note: widgets have their own visibility controls already, so even if a page is visible that doesn't mean that all widgets associated with the page will be as well.

gmcauliffe: That's great. Pages should have their own visibility. Thanks for checking 👍

bronson-c (approved):
Just one minor change that would be nice to have, looks great!

firstvoices/backend/models/page.py
```diff
@@ -0,0 +1,65 @@
...
+            "add": predicates.is_superadmin,  # permissions will change when we add a write API
+            "change": predicates.is_superadmin,
+            "delete": predicates.is_superadmin,
+        }
+        constraints = [
+            # Constraint to ensure a SitePage has either a banner image or video or none but not both.
```
sarahfirstvoices, line 24: Is there a way to reuse this between SitePage and Site?

firstvoices/backend/models/page.py
```diff
@@ -0,0 +1,65 @@
...
+                    Q(
+                        Q(banner_image=None, banner_video=None)
+                        | Q(Q(banner_image=None) & ~Q(banner_video=None))
+                        | Q(~Q(banner_image=None) & Q(banner_video=None))
+                    )
+                ),
+            )
+        ]
+
+    title = models.CharField(max_length=225)
```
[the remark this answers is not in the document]
sarahfirstvoices: @gmcauliffe and maybe @caforbes might have ideas here. Do we want to have the same limit for any other titles? Maybe pages, songs, and stories could have the same title length? Or should this limit be just for page titles?

firstvoices/backend/models/page.py
```diff
@@ -0,0 +1,65 @@
...
+    slug = models.SlugField(
+        max_length=200,
+        blank=False,
+        validators=[validate_slug],
+        db_index=True,
+        unique=True,
+    )
```
sarahfirstvoices, line 44: @gmcauliffe Do we currently restrict the page urls at all, to skip reserved ones like "dictionary" etc?

firstvoices/backend/models/page.py
```diff
@@ -0,0 +1,65 @@
...
+    subtitle = models.TextField(blank=True)
```
sarahfirstvoices, line 45: Should this maybe be a CharField with a max limit?

firstvoices/backend/models/page.py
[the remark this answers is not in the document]
gmcauliffe: I think pages, songs, and stories could all have the same max length for title 👍

firstvoices/backend/tests/test_models/test_page_model.py
```diff
@@ -0,0 +1,24 @@
...
+    @pytest.mark.django_db
+    def test_one_or_zero_banner(self):
+        image = factories.ImageFactory.create()
+        video = factories.VideoFactory.create()
+
+        try:
+            factories.SitePageFactory.create()
+            factories.SitePageFactory.create(banner_image=image)
+            factories.SitePageFactory.create(banner_video=video)
+        except IntegrityError:
```
sarahfirstvoices, line 17: Could you adjust this a little bit to be sure the error is coming when you try to add the second banner and not before?

firstvoices/backend/models/page.py
gmcauliffe: There is a custom page route "sandbox" of sorts in the frontend to prevent conflicts with any of our FirstVoices paths like "dictionary".  A list of reserved paths in the BE would be great (if it isn't too time consuming) and would help to prevent any potential confusion caused by duplicates like this:
"smalgyax/dictionary" and "smalgyax/custom-page/dictionary.

On a separate point, We definitely want paths to be unique within a site but NOT unique across all sites (i.e. two different language sites should be able to have the same path e.g. "smalgyax/language-tools" and "SECOTEN/language-tools").

Thanks for raising this :)

firstvoices/backend/models/page.py
NolanVH: It looks like constraints are not reusable as they require a unique name across models. I have updated the constraint to match the one in the site model though.

firstvoices/backend/models/page.py
NolanVH: Thanks for the feedback!

I have added a [`unique_together`](https://github.com/First-Peoples-Cultural-Council/fv-be/blob/b3d0a7b33bba2293b8dd4394d59bdddef0e4c3ba/firstvoices/backend/models/page.py#L31) Meta option to the model which will ensure that the site page slugs are unique across a site but can be the same on two different sites. 

I have also [created a validator](https://github.com/First-Peoples-Cultural-Council/fv-be/blob/b3d0a7b33bba2293b8dd4394d59bdddef0e4c3ba/firstvoices/backend/models/validators.py#L1) that checks that site page slugs do not contain any of the strings in the reserved strings list. @gmcauliffe would you be able to point me to the current frontend list of reserved paths so I can match them up? If there are any others that we might want to add I can do that as well.

Another option here would be to store reserved slug strings in a model that can be updated without pushing code changes. If others have thoughts about this I am open to suggestions. @sarahfirstvoices

sarahfirstvoices: I think listing the reserved slugs here is great for mvp, and we can consider making it configurable in the future.

sarahfirstvoices (approved):
Looks good to me! Thanks, Nolan

PR #332, opened 2023-07-26
FW-4216 Add SitePage create/update/delete APIs

NolanVH:
### Description of Changes
This PR adds the SitePage create/update/delete APIs.

The create and update APIs are passed a list of SiteWidget IDs and the serializer automatically creates or updates a SiteWidgetList instance, as well as the SiteWidgetListOrder instances based on the order that the widgets are passed to the API. The delete API automatically deletes the SiteWidgetList and SiteWidgetListOrder objects associated with the SitePage. Tests have been added ensuring that the related models are correctly created or deleted as required.

### Checklist

### Additional Notes
N/A


firstvoices/backend/serializers/page_serializers.py
```diff
@@ -38,3 +47,62 @@ class Meta(SiteContentLinkedTitleSerializer.Meta):
...
             "banner_video",
         )
+
+
+class SitePageDetailWriteSerializer(
+    UpdateSerializerMixin, CreateSiteContentSerializerMixin, SitePageDetailSerializer
+):
+    widgets = serializers.PrimaryKeyRelatedField(
+        queryset=SiteWidget.objects.all(),
+        many=True,
```
sarahfirstvoices, line 57: Should widgets use the SameSite() validator?

firstvoices/backend/serializers/page_serializers.py
```diff
@@ -38,3 +47,62 @@ class Meta(SiteContentLinkedTitleSerializer.Meta):
...
+    banner_video = serializers.PrimaryKeyRelatedField(
+        queryset=Video.objects.all(),
+        allow_null=True,
+        validators=[SameSite()],
+    )
+
+    def validate_widgets(self, widgets):
+        if not widgets:
+            raise serializers.ValidationError(
+                "The widgets field must contain at least one widget ID."
```
sarahfirstvoices, line 73: Is this too constrained? Would it be useful for people to save a page with just a banner and title, since they can make it team-only?

firstvoices/backend/serializers/page_serializers.py
```diff
@@ -38,3 +47,62 @@ class Meta(SiteContentLinkedTitleSerializer.Meta):
...
+                    f"SiteWidget with ID ({site_widget.id}) does not belong to the site."
+                )
+        return widgets
+
+    def to_representation(self, instance):
+        data = SitePageDetailSerializer(instance=instance, context=self.context).data
+        return data
+
+    def create(self, validated_data):
+        validated_data["visibility"] = Visibility[
```
sarahfirstvoices, line 89: Should we pull visibility handling into a mixin for all the controlled site content?

firstvoices/backend/tests/test_apis/base_api_test.py
```diff
@@ -491,7 +493,9 @@ def test_update_403(self):
         instance = self.create_minimal_instance(site=site, visibility=Visibility.PUBLIC)
 
         response = self.client.put(
-            self.get_detail_endpoint(key=instance.id, site_slug=site.slug),
+            self.get_detail_endpoint(key=instance.slug, site_slug=site.slug)
+            if instance.__class__ == SitePage
+            else self.get_detail_endpoint(key=instance.id, site_slug=site.slug),
```
sarahfirstvoices, line 498: Let's not put a class-specific condition in the base test class. Suggested other way to do this:

self.get_detail_endpoint(key=self.get_lookup_key(instance), ....

Add a default get_lookup_key method that returns instance.id, and override that in the site page tests.

firstvoices/backend/tests/test_apis/test_pages_api.py
```diff
@@ -2,26 +2,81 @@
...
+            "banner_image": str(banner_image.id),
+            "banner_video": None,
+        }
+
+    def add_related_objects(self, instance):
+        # no related objects to add
+        pass
+
+    def assert_related_objects_deleted(self, instance):
+        # no related objects to delete
```
sarahfirstvoices, line 50: The list order items should be deleted when the page is deleted, right? This would be a good place to assert that

firstvoices/backend/tests/test_apis/base_api_test.py
NolanVH: Great suggestion, thanks!

firstvoices/backend/serializers/page_serializers.py
NolanVH: I have added a mixin to do this. Thanks.

firstvoices/backend/models/page.py
```diff
@@ -40,7 +40,7 @@ class Meta:
     )
     subtitle = models.CharField(blank=True, max_length=constants.DEFAULT_TITLE_LENGTH)
     widgets = models.ForeignKey(
-        SiteWidgetList, on_delete=models.CASCADE, related_name="sitepage_set"
+        SiteWidgetList, on_delete=models.PROTECT, related_name="sitepage_set"
```
sarahfirstvoices, line 43: Should this be SET_NULL instead?

firstvoices/backend/serializers/base_serializers.py
```diff
@@ -84,3 +85,16 @@ class CreateSiteContentSerializerMixin(CreateSerializerMixin):
     def create(self, validated_data):
         validated_data["site"] = get_site_from_context(self)
         return super().create(validated_data)
+
+
+class CreateSiteContentVisibilitySerializerMixin(CreateSiteContentSerializerMixin):
```
sarahfirstvoices, line 90: Maybe CreateControlledSiteContentSerializerMixin, to match how the base models are named?

firstvoices/backend/serializers/page_serializers.py
```diff
@@ -1,18 +1,26 @@
...
 
 
 class SitePageSerializer(SiteContentLinkedTitleSerializer):
     url = SiteHyperlinkedIdentityField(
-        view_name="api:sitepage-detail", lookup_field="slug"
+        view_name="api:sitepage-detail", lookup_field="slug", read_only=True
     )
-    slug = serializers.CharField(read_only=True)
-    visibility = serializers.CharField(read_only=True, source="get_visibility_display")
+    slug = serializers.CharField()
```
sarahfirstvoices, line 22: Did you check this one with @caforbes ? Not sure we want to let teams edit the slug for their site

firstvoices/backend/views/page_views.py
```diff
@@ -59,10 +65,83 @@
...
+                    fields={
+                        "id": serializers.UUIDField(),
+                        "title": serializers.CharField(),
+                        "url": serializers.URLField(),
+                        "visibility": serializers.CharField(),
+                        "subtitle": serializers.CharField(),
+                        "slug": serializers.CharField(),
+                        "widgets": SiteWidgetDetailSerializer(many=True),
+                        "banner_image": ImageSerializer(),
+                        "banner_video": VideoSerializer(),
```
sarahfirstvoices, line 85: Just noticed these. Is there a reason we can't just refer to the existing serializer here? Same question for the other inline serializers as well

sarahfirstvoices (approved):
Approved, if Clarissa approves of the editable slugs

firstvoices/backend/models/page.py
NolanVH: The reasoning for having PROTECT here is that in the admin panel if a user tries to delete a SiteWidgetList they will get an error if it is being used by a page. As far as the page APIs go it would work either way and I would be happy to change it.

firstvoices/backend/views/page_views.py
NolanVH: If the `SitePageDetailSerializer` is used here than the widgets field formats incorrectly since the `SitePageDetailSerializer` uses the `SiteWidgetListSerializer` for the widgets field (which correctly creates the needed model instances) instead of the `SiteWidgetDetailSerializer` (which is the needed serializer for the API docs display).

In other words it looks like this:
Screenshot 2023-07-26 at 3 33 13 PM

But should look like this:
Screenshot 2023-07-26 at 3 33 43 PM

firstvoices/backend/serializers/page_serializers.py
caforbes: Sarah is right, both the language site slugs and the page slugs should not be editable by teams. In the frontend forms it is explicitly said "you cannot edit this".

NolanVH: Good catch, thanks! I will update the PR.

PR #381, opened 2023-08-09
Fix media files not deleting from AWS using Admin

NolanVH:
### Description of Changes
This PR fixes the media files not being removed from AWS when deleting from the admin list view (eg: from `/admin/backend/imagefile/`). When media is deleted this way `delete_queryset` is called which doesn't use the model `delete` method, leaving some orphaned media on AWS in certain cases.

Relates to [FW-4547](https://firstvoices.atlassian.net/browse/FW-4547), [FW-4352](https://firstvoices.atlassian.net/browse/FW-4352), and [FW-4676](https://firstvoices.atlassian.net/browse/FW-4676).

### Checklist

### Additional Notes
N/A

[FW-4547]: https://firstvoices.atlassian.net/browse/FW-4547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FW-4352]: https://firstvoices.atlassian.net/browse/FW-4352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FW-4676]: https://firstvoices.atlassian.net/browse/FW-4676?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/firstvoices/backend/admin/media_admin.py
+++ b/firstvoices/backend/admin/media_admin.py
@@ -22,6 +22,10 @@ class FileAdmin(HiddenBaseAdmin):
     search_fields = ("content",)
     readonly_fields = ("mimetype", "size") + HiddenBaseAdmin.readonly_fields
 
+    def delete_queryset(self, request, queryset):
+        for obj in queryset.all():
+            obj.delete()
+
 
 @admin.register(ImageFile)
 @admin.register(VideoFile)
@@ -37,6 +41,10 @@ class AudioAdmin(BaseSiteContentAdmin):
     list_display = ("title",) + BaseSiteContentAdmin.list_display
     search_fields = ("title",)
 
+    def delete_queryset(self, request, queryset):
+        for obj in queryset.all():
+            obj.delete()
+
 
 @admin.register(Image)
 @admin.register(Video)
@@ -48,6 +56,10 @@ class VisualMediaAdmin(BaseSiteContentAdmin):
     ) + BaseSiteContentAdmin.readonly_fields
     list_display = ("title",) + BaseSiteContentAdmin.list_display
     search_fields = ("title",)
+
+    def delete_queryset(self, request, queryset):
+        for obj in queryset.all():
+            obj.delete()
 
 
 @admin.register(EmbeddedVideo)


caforbes (approved):
Awesome catch!

sarahfirstvoices:
Is this going to prevent us from bulk-removing media models while leaving the files intact, during migration testing? Should we maybe wait until after migrating our prod data to merge this?

caforbes:
I think we can leave the File objects (and AWS files) intact. I was imagining this workflow:

- As the first step of migration, unlink File objects from the Audio/Image/Video objects on affected sites.
- Then delete the affected sites prior to migration as usual - this will not delete the Files since they are unlinked
- As part of the migration, create the media objects
- Pass the AWS path to those objects; if there is a File object with that path, get it and link it to the media object
- If there is no File object with that AWS path, then create it with raw SQL insertion, then get it and link it to the media object

That way we can have fixed files in AWS regardless of environment, create the File objects in the db where needed and let them persist across migration iterations without needing to call File.delete at any point, from the admin panel or otherwise. We can add a step for dangling Files if needed. Does that make sense/work?

NolanVH:
@sarahfirstvoices we would still be able to remove media models without deleting the files from AWS by deleting the file instances from the database directly (backend_imagefile, backend_videofile, and backend_file tables). We could also leave this PR unmerged and use the admin list view delete to accomplish the same and create a ticket to merge this later.

@caforbes The file instances all have foreign keys to a site with on_delete set to cascade, so the file objects would get deleted when we delete the sites. It would also take some processing to check if a file instance with the correct path exists when we import media.

I had imagined that we would do the following:
- delete the file objects either by deleting the rows in the database tables directly or by leaving this PR unmerged and deleting them in the admin panel (this would leave the AWS files and paths intact).
- delete the media objects and sites as needed prior to a fresh import
- create the media objects during the import
- during each media instance creation we also create the file instance by directly inserting a row into the database and link the file instance to the media instance since they are one-to-one
    - this row would contain the AWS path which can be constructed with import data

One thing I wondered about is that the existing media objects in Nuxeo have a single UUID. Would it make more sense to use this UUID on the file instances or on the media instances? My thought is that it might make sense to use it on the file instances so that it would be easy to add the related key to the media instances.

caforbes:
The media UUID needs to be used on the media models because all other site content (e.g. dictionaryentry) references that id. could we use the same id on both Media and File objects for now? I had imagined that we could create new ids for the File objects and find the correct ones based on the AWS path, which can be generated from other media model info.

NolanVH:
We can't use the same ID for both media and file models since they have to be unique across the platform but it should be easy enough to create a new UUID for the file models and link them to the media models during the import step.

PR #402, opened 2023-08-16
FW-4354 import memberships

caforbes:
### Description of Changes
Adds import for app-level and site-level memberships. Skips users with existing app-level memberships (e.g. don't overwrite existing superadmins). Also fixes a bug that appears when re-migrating users -- skip import for users already in the db.

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/resources/utils/import_export_widgets.py
+++ b/firstvoices/backend/resources/utils/import_export_widgets.py
@@ -1,6 +1,8 @@
 from django.contrib.auth import get_user_model
 from import_export.widgets import ForeignKeyWidget, Widget
 
+DUMMY_USER_EMAIL = "test@test.com"
+
 
 class ChoicesWidget(Widget):
     """Import/export widget to use choice labels instead of internal value.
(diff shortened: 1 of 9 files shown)


firstvoices/backend/resources/utils/import_export_widgets.py
```diff
@@ -29,29 +31,23 @@ def render(self, value, obj=None):
 class UserForeignKeyWidget(ForeignKeyWidget):
     """Import/export widget to find/create users from their email.
 
-    When Django cannot find a User with a matching email address,
-    it will create a new one if create=True, otherwise, will return dummy user.
+    If Django cannot find a User with a matching email address, will return dummy user.
```
sarahfirstvoices, line 34: Is there a reason not to create a user for unknown email addresses that show up in the created_by/modified_by fields?

firstvoices/backend/resources/utils/import_export_widgets.py
```diff
@@ -1,6 +1,8 @@
 from django.contrib.auth import get_user_model
 from import_export.widgets import ForeignKeyWidget, Widget
 
+DUMMY_USER_EMAIL = "test@test.com"
```
sarahfirstvoices, line 4: This might eventually be visible to team members-- should we maybe use something like support@fpcc.ca as the dummy account?

sarahfirstvoices (approved):

firstvoices/backend/resources/utils/import_export_widgets.py
caforbes: I had made this option but we were not actually using it anywhere. Should I revert the change and set the migrations to create users? I had thought that any users that were not migrated (e.g. `jenkins_cli`) would be better handled with the dummy user, once it looks like an administrator.

PR #410, opened 2023-08-18
Fw 4696 v 3 be wotd pulling from non public words also phrases

sahil97:
### Description of Changes
- Added visibility matching for word of the day entries
- Added permissions to dictionary entries on word-of-the-day entries before returning

### Checklist

### Additional Notes
N/A


firstvoices/backend/views/word_of_the_day_views.py
```diff
@@ -140,10 +146,19 @@ def get_selected_word(self, site_slug):
...
+
+        if selected_word:
+            unfiltered_queryset = DictionaryEntry.objects.filter(
+                id=selected_word[0].dictionary_entry.id
+            )
+            filtered_queryset = utils.filter_by_viewable(
+                request.user, unfiltered_queryset
+            )
+            if filtered_queryset:
+                queryset = selected_word
```
NolanVH, line 161: Won't this just return no word of the day if the user doesn't have the ability to see the selected word of the day and is that the functionality we want? I'm just wondering if a different word of the day should be selected if a user doesn't have permission to see the current word.

NolanVH (approved):
Aside from the one question I had, all changes look good to me!

firstvoices/backend/views/word_of_the_day_views.py
sahil97: Yes, that is by design. A few cases can arise where the visibility of the updated word does not allow it to be viewed by the user, but since the word of the day can only be chosen once per date, we cannot select/show an alternative word and thus no word is returned.

PR #412, opened 2023-08-18
FW-4445 SitePage data import

NolanVH:
### Description of Changes
This PR adds SitePage data imports. The SitePage model now has the widgets field set to allow null since some of the pages on Nuxeo have no widgets.

### Checklist

### Additional Notes
N/A


bronson-c (approved):
LGTM! 👍

firstvoices/backend/resources/pages.py
```diff
@@ -0,0 +1,73 @@
...
+        # Create a SiteWidgetList object for each SitePage and convert the list of widget IDs to SiteWidgetListOrder
+        # objects connecting the list to the widgets.
+        if row["widgets"] == "":
+            row["widgets"] = None
+        else:
+            last_modified_by_user = self.get_user_or_none(row["last_modified_by"])
+            created_by_user = self.get_user_or_none(row["created_by"])
+            site = Site.objects.get(id=row["site"])
+            widgets_list = row["widgets"].split(",")
+            site_widget_list = SiteWidgetList.objects.create(site=site)
```
caforbes, line 43: Just wondering if these object creation steps could potentially cause an error or some orphans if we imported with `dry_run=True`? Or should we just commit to avoid that?

firstvoices/backend/resources/pages.py
```diff
@@ -0,0 +1,73 @@
...
+                logger.warning(
+                    f"Image (banner) with ID ({row['banner_image']}) could not be found when creating SitePage with ID "
+                    f"({row['id']})."
+                )
+                row["banner_image"] = None
+
+        if row["banner_video"] and row["banner_video"] != "":
+            try:
+                Video.objects.get(id=row["banner_video"])
+            except Image.DoesNotExist:
```
caforbes, line 68: Does this need to be `Video.DoesNotExist`?

caforbes (approved):
Two small questions, which could be non-issues or quick fixes.

firstvoices/backend/resources/pages.py
NolanVH: I believe if we run with `dry_run=True` almost every import will fail since they all depend on some other table (mainly the sites). I think we can probably leave this as is and not use `dry_run=True` but I am open to other suggestions.

PR #453, opened 2023-08-29
FW-4775 Add site homepage widget data migration

NolanVH:
### Description of Changes
This PR adds site homepage widget data migrations. This is done in a separate resource from the original site import while still using the same site CSV. This is because the sites must be imported, followed by the widgets, followed by the site homepage field.

Additionally, the created/modified fields have been added to the SitePage data migration.

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/management/commands/import_csv_data.py
+++ b/firstvoices/backend/management/commands/import_csv_data.py
@@ -133,14 +133,8 @@ def run_import():
 
     for key, resource in import_resources:
         # Parse files to import with this resource
-        if resource.__class__ == SiteHomepageWidgetsResource:
-            logger.info("Importing Site homepage widget lists...")
-            matched_files = [
-                f for f in os.listdir(current_export_dir) if f.startswith(key)
-            ]
-        else:
-            logger.info(f"Importing {key} models...")
-            matched_files = [f for f in unmatched_files if f.startswith(key)]
+        logger.info(f"Importing from [{key}] CSV with {type(resource).__name__}...")
+        matched_files = [f for f in unmatched_files if f.startswith(key)]
         unmatched_files = [f for f in unmatched_files if f not in matched_files]
 
         if not matched_files:
--- a/firstvoices/backend/management/commands/import_csv_data.py
+++ b/firstvoices/backend/management/commands/import_csv_data.py
@@ -35,6 +35,7 @@ from backend.resources.media import (
     VideoResource,
 )
 from backend.resources.pages import SitePageResource
+from backend.resources.site_homepage_widgets import SiteHomepageWidgetsResource
 from backend.resources.sites import (
     MembershipResource,
     SiteMigrationResource,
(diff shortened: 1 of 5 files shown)


firstvoices/backend/management/commands/import_csv_data.py
```diff
@@ -133,7 +135,12 @@ def run_import():
         logger.info(f"Importing {key} models...")
 
         # Parse files to import with this resource
-        matched_files = [f for f in unmatched_files if f.startswith(key)]
+        if resource.__class__ == SiteHomepageWidgetsResource:
+            matched_files = [
+                f for f in os.listdir(current_export_dir) if f.startswith(key)
+            ]
+        else:
+            matched_files = [f for f in unmatched_files if f.startswith(key)]
```
NolanVH, line 143: The purpose of this is to reuse the same sites CSV file for the site homepage widgets import.

firstvoices/backend/resources/pages.py
NolanVH, line 1: Added some missing created/modified fields to the pages import.

firstvoices/backend/resources/pages.py
```diff
@@ -40,15 +40,23 @@ def before_import_row(self, row, **kwargs):
             created_by_user = self.get_user_or_none(row["created_by"])
             site = Site.objects.get(id=row["site"])
             widgets_list = row["widgets"].split(",")
-            site_widget_list = SiteWidgetList.objects.create(site=site)
+            site_widget_list = SiteWidgetList.objects.create(
```
bronson-c, line 43: When running locally I get warnings for each SiteWidgetList and SiteWidgetListOrder model created in this `before_import_row` function.

e.g.  RuntimeWarning: DateTimeField SiteWidgetList.last_modified received a naive datetime (2022-09-12 14:38:27.007000) while time zone support is active.

Would it be worth it to convert these date values to proper ones including a timezone or can the warnings be ignored? CC: @caforbes

bronson-c: Thes warnings occur in both the page import and site_homepage_widget import

firstvoices/backend/management/commands/import_csv_data.py
bronson-c: [minor] I'm on board with re-using the sites csv, but in the logs when the site homepage widgets are being imported the log states:
INFO - Importing sites models...
INFO - Importing file: sites_[date].csv

Which may be confusing when importing real data. I would recommend overriding the logs ffor this specific resource as well

firstvoices/backend/resources/pages.py
NolanVH: Good catch! The timestamps could probably be removed from the import and possibly added into other imports as a mixin later. @caforbes do you have any thoughts on this?

caforbes: After thinking about it, I don't think it is really important to preserve any created/last_modified data for this intermediate model which does not have a nuxeo correlate. It's completely fine if the created/last_modified dates default to the migration time.

caforbes (approved):
Tested locally and looks good, runs without warnings.

PR #463, opened 2023-09-01
FW-4820 Subtitle should not be required

bronson-c:
### Description of Changes
Adds the `NullableCharField` to convert null values to empty strings and adds the field to page subtitles. This field can be re-used in future tickets.

### Checklist

### Additional Notes
The 400 error was caused by the create page form supplying `null` to the page create API "subtitle" field rather than an empty string. If for some reason we'd want to disallow null values in this field, the issue could also be fixed by editing what the create page form sends.


gmcauliffe (approved):
Thank you! LGTM 👍

sahil97 (approved):
LGTM !

firstvoices/backend/models/page.py
```diff
@@ -38,7 +38,9 @@ class Meta:
         db_index=True,
         unique=False,
     )
-    subtitle = models.CharField(blank=True, max_length=constants.DEFAULT_TITLE_LENGTH)
+    subtitle = models.CharField(
+        null=True, blank=True, max_length=constants.DEFAULT_TITLE_LENGTH
+    )
```
sarahfirstvoices, line 43: Please don't allow nulls in CharFields: https://docs.djangoproject.com/en/4.2/ref/models/fields/#null

If we want to allow the frontend to send null values, we need to handle that in the serializer instead (convert null to "" when saving). I'll log a ticket to do that for all string fields in all the serializers eventually, but feel free to set it up for this one API as part of this ticket, or just ask the frontend to send "" for now

firstvoices/backend/tests/test_apis/test_pages_api.py
```diff
@@ -90,7 +90,7 @@ def get_expected_response(self, instance, site):
         ] = f"http://testserver{self.get_detail_endpoint(key=instance.slug, site_slug=site.slug)}"
         return {
             **controlled_standard_fields,
-            "subtitle": "",
+            "subtitle": None,
```
sarahfirstvoices, line 93: Instead of changing this, please add a new test to check support for the null value

firstvoices/backend/tests/test_apis/test_pages_api.py
```diff
@@ -103,7 +103,7 @@ def get_expected_detail_response(self, instance, site):
...
```
sarahfirstvoices, line 106: Same comment about adding a second test (or parametrize these)

sarahfirstvoices (changes requested):
Please don't merge this. We don't want null string values in the db

firstvoices/backend/serializers/fields.py
```diff
@@ -88,3 +88,10 @@ def to_representation(self, value):
             return visibility_map[value]
         except KeyError:
             raise serializers.ValidationError("Invalid visibility value.")
+
+
+class NonNullCharField(serializers.CharField):
```
sarahfirstvoices, line 93: This is a great solution and implementation. Nice!

My only quibble is about the name. This is a CharField that allows null input, so maybe NullableCharField makes more sense?

sarahfirstvoices (approved):

firstvoices/backend/serializers/fields.py
sarahfirstvoices: No need to change it now, and no need to re-review if you do!

PR #475, opened 2023-09-07
FW-4234 improved admin forms

caforbes:
### Description of Changes

Added improved site-by-site filtering for content on misc admin forms, with particular attention to the Site form and Character forms (addressing some of FW-4234). There is now a site filter in various admin lists. Dropdowns now only contain content from the same site (this fixes WOTD bug FW-4405). The filtered queries improve admin site performance/load times.

### Checklist
- Tests have been updated / created if applicable

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/admin/characters_admin.py
+++ b/firstvoices/backend/admin/characters_admin.py
@@ -9,7 +9,6 @@ from backend.models.characters import (
 from backend.models.dictionary import DictionaryEntry
 
 from .base_admin import (
-    BaseAdmin,
     BaseInlineAdmin,
     BaseInlineSiteContentAdmin,
     BaseSiteContentAdmin,
@@ -111,9 +110,7 @@ class IgnoredCharacterAdmin(BaseSiteContentAdmin):
 
 
 @admin.register(Alphabet)
-class AlphabetAdmin(BaseAdmin):
+class AlphabetAdmin(BaseSiteContentAdmin):
     fields = ("site", "input_to_canonical_map")
-    list_display = (
-        "site",
-        "input_to_canonical_map",
-    ) + BaseAdmin.list_display
+    list_display = ("input_to_canonical_map",) + BaseSiteContentAdmin.list_display
+    list_filter = ()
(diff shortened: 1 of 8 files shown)


firstvoices/backend/admin/characters_admin.py
```diff
@@ -74,37 +111,9 @@ class IgnoredCharacterAdmin(BaseSiteContentAdmin):
 
 
 @admin.register(Alphabet)
-class AlphabetAdmin(BaseSiteContentAdmin):
+class AlphabetAdmin(BaseAdmin):
```
sarahfirstvoices, line 114: Why the change in base class? It's minor but the BaseSiteContentAdmin does a select_related for the site field that helps when there are a lot of items on the list page

firstvoices/backend/admin/sites_admin.py
```diff
@@ -78,9 +78,9 @@ class SiteFeatureAdmin(BaseSiteContentAdmin):
 
 
 @admin.register(SiteMenu)
-class SiteMenuAdmin(BaseSiteContentAdmin):
+class SiteMenuAdmin(BaseAdmin):
     fields = ("site", "json")
-    list_display = ("json",) + BaseSiteContentAdmin.list_display
+    list_display = ("json", "site") + BaseAdmin.list_display
```
sarahfirstvoices, line 83: Same question: any reason for this change?

firstvoices/backend/admin/characters_admin.py
caforbes: I changed it (for both) because these models essentially provide 1 object per site, and the only content is a json field. I had updated BaseSiteContentAdmin to let you filter rows per site, and that wasn't necessary when there is only 1 row per site (or less) in the first place. An alternative would be to return this to BaseSiteContentAdmin and remove the filter and duplicate `site` columns, if that would be better.

sarahfirstvoices: Oh, gotcha! I vote for that alternate way, just to make sure we get all the prefetching everywhere

sarahfirstvoices: I'm marking this approved because I think it's fine either way. No need to re-review if you make that change :)

sarahfirstvoices (approved):

PR #489, opened 2023-09-11
FW-4913 Fix site creation via admin site

NolanVH:
### Description of Changes
This PR fixes the small bug when creating a site on the admin forms.

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/admin/base_admin.py
+++ b/firstvoices/backend/admin/base_admin.py
@@ -154,7 +154,7 @@ class BaseInlineSiteContentAdmin(BaseInlineAdmin):
         if not hasattr(self, "parent_site"):
             self.parent_site = None
             object_id = request.resolver_match.kwargs.get("object_id")
-            instance = self.parent_model.objects.filter(id=object_id)
+            instance = self.parent_model.objects.filter(id=object_id).first()
             if instance and isinstance(instance, Site):
                 self.parent_site = instance
             elif instance and hasattr(instance, "site"):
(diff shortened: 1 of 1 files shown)


firstvoices/backend/admin/base_admin.py
```diff
@@ -154,7 +154,7 @@ def get_site_from_object(self, request):
         if not hasattr(self, "parent_site"):
             self.parent_site = None
             object_id = request.resolver_match.kwargs.get("object_id")
-            instance = self.parent_model.objects.get(id=object_id)
+            instance = self.parent_model.objects.filter(id=object_id)
```
sahil97, line 157: Would this not return a `queryset/list` ? and probably both 'if' clauses below this wouldn't work as they are both testing the `instance` type but the type would always be a `queryset/list` ?

firstvoices/backend/admin/base_admin.py
```diff
@@ -154,7 +154,7 @@ def get_site_from_object(self, request):
...
             if instance and isinstance(instance, Site):
```
sarahfirstvoices, line 158: Does this condition ever find a Site? I would expect it to always be a queryset. I think we need to add .first() in there somewhere

sahil97: Same timing. Haha

firstvoices/backend/admin/base_admin.py
NolanVH: Good call, thanks! I have updated this to return the first instance.

firstvoices/backend/admin/base_admin.py
NolanVH: Good call, thanks! I have updated this to return the first instance.

sahil97 (approved):
LGTM !

sarahfirstvoices (approved):

PR #493, opened 2023-09-12
FW-4888 Categories search index fixes

NolanVH:
### Description of Changes
This PR adds the following three changes:
- Refreshes the dictionary entry/song/story indices when a dictionary entry is saved to allow related field signals to grab the index before the default 1 second has passed between ES updates.
- Adds a category field index update on the m2m_changed signal which is used in the APIs.
- Includes parent category IDs in the index so that dictionary entries will be returned when filtering entries with child categories by a parent category.

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/search/indices/dictionary_entry_document.py
+++ b/firstvoices/backend/search/indices/dictionary_entry_document.py
@@ -1,7 +1,8 @@
 import logging
 
-from django.db.models.signals import post_delete, post_save
+from django.db.models.signals import m2m_changed, post_delete, post_save
 from django.dispatch import receiver
+from django_elasticsearch_dsl import Index
 from elasticsearch.exceptions import ConnectionError, NotFoundError
 from elasticsearch_dsl import Keyword, Text
 
@@ -93,6 +94,9 @@ def update_dictionary_entry_index(sender, instance, **kwargs):
                 visibility=instance.visibility,
             )
             index_entry.save()
+        # Refresh the index to ensure the index is up-to-date for related field signals
+        dict_index = Index(ELASTICSEARCH_DICTIONARY_ENTRY_INDEX)
+        dict_index.refresh()
     except ConnectionError as e:
         logger = logging.getLogger(ELASTICSEARCH_LOGGER)
         logger.error(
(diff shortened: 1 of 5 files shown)


sahil97 (approved):
Tested locally and works succesfully. M2M and foreign key fields are being added to the indexed document. 
LGTM !

firstvoices/backend/search/utils/object_utils.py
```diff
@@ -171,8 +172,15 @@ def get_notes_text(dictionary_entry_instance):
...
-        for id in dictionary_entry_instance.categories.values_list("id", flat=True)
+        str(category_id)
+        for category_id in list(
+            chain(
+                dictionary_entry_instance.categories.values_list("id", flat=True),
+                dictionary_entry_instance.categories.values_list(
+                    "parent_id", flat=True
+                ),
+            )
+        )
```
sarahfirstvoices, line 183: Rather than saving the parent category id in the es index, it would probably be better to update the search api to query for all the category IDs it needs. If we story the parent category id in the index, that means we need to update potentially large amounts of dictionary entries in the index if someone edits a Category instance.

firstvoices/backend/search/indices/dictionary_entry_document.py
```diff
@@ -271,14 +274,25 @@ def update_acknowledgement(sender, instance, **kwargs):
...
+
+# Category update when called through the APIs
+@receiver(m2m_changed, sender=DictionaryEntryCategory)
+def update_categories_m2m(sender, instance, **kwargs):
+    if instance.__class__ == DictionaryEntry:
+        dictionary_entry = instance
+        update_dictionary_entry_index_categories(dictionary_entry, instance)
 
+
+def update_dictionary_entry_index_categories(dictionary_entry, instance):
```
sarahfirstvoices, line 293: Why do we need both these parameters, if they are the same?

NolanVH: Good catch. I have updated this.

firstvoices/backend/search/utils/object_utils.py
NolanVH: Yes, good call. I have removed the parent category ID from the index and have checked that the functionality is now correct. 

When a child ID is given to the search API it will return only entries that have the child ID. If a parent ID is given, the API will return entries that have either the parent or a child category.

sarahfirstvoices (approved):

PR #515, opened 2023-09-18
FW-4938 Audio list DB performance improvements

NolanVH:
### Description of Changes
These changes improve the Audio list view database query performance as follows:
- The query to get the pagination object count has been optimized to fetch only the ID instead of all fields on an object reducing the query latency by approximately 75%.
- The get_queryset function in AudioViewSet has been updated to use `prefetch_related` for the site and original relations instead of `select_related`, reducing the query latency by approximately 80%.
- A few of the unneeded fields have been removed from the query selection using `defer`.

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/pagination.py
+++ b/firstvoices/backend/pagination.py
@@ -15,10 +15,10 @@ class FasterCountPagination(Paginator):
     @cached_property
     def count(self):
         # Override the count property to select only the id field to speed up the count query
-        c = getattr(self.object_list.values("id"), "count", None)
+        c = getattr(self.object_list.values("pk"), "count", None)
         if callable(c) and not inspect.isbuiltin(c) and method_has_no_args(c):
             return c()
-        return len(self.object_list.values("id"))
+        return len(self.object_list.values("pk"))
 
 
 class PageNumberPagination(pagination.PageNumberPagination):
--- a/firstvoices/backend/pagination.py
+++ b/firstvoices/backend/pagination.py
@@ -1,7 +1,9 @@
+import inspect
 from collections import OrderedDict
 
 from django.core.paginator import InvalidPage, Page, Paginator
 from django.utils.functional import cached_property
+from django.utils.inspect import method_has_no_args
 from rest_framework import pagination
 from rest_framework.exceptions import NotFound
 from rest_framework.response import Response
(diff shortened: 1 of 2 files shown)


sahil97 (approved):
LGTM !

firstvoices/backend/pagination.py
```diff
@@ -1,16 +1,25 @@
...
 from rest_framework.response import Response
 
 from backend.search.utils.constants import ES_MAX_RESULTS, ES_PAGE_SIZE
 
 
+class FasterCountPagination(Paginator):
+    @cached_property
+    def count(self):
+        # Override the count property to select only the id field to speed up the count
+        return self.object_list.values("id").count()
```
sarahfirstvoices, line 16: I am super into this optimization, and I have a couple of thoughts.

1. This is worth contributing back to Django core.
2. But, this skips some of the safety checks from the original method and they should be added back in. (I believe they are there to make this work for both querysets and regular lists.) If you don't want to test that right now, feel free to log a separate ticket to flesh it out, since we currently only use the paginator for querysets

sarahfirstvoices: And a question: is there any possibility that doing this produces a different count than the original query, especially if there are table joins without using distinct()? Have you tested it?

NolanVH: Some great feedback, thanks!

1. I believe the Django core is this way because it ensures compatibility with all models, not just applications that use a UUID as a primary key.
2. Great point. I will add that in.

I have tested and the sizes come back as the same. The difference between the queries is that the original one is doing the following: 
```
SELECT COUNT(*)
  FROM (
        SELECT DISTINCT "backend_audio"."id" AS "col1",
               "backend_audio"."created" AS "col2",
               "backend_audio"."site_id" AS "col3",
               ...,
          FROM "backend_audio"...
```
and the updated one is doing the following:
```
SELECT COUNT(*)
  FROM (
        SELECT DISTINCT "backend_audio"."id" AS "col1"
          FROM "backend_audio"...
```
The joins on site and membership are the same in both queries.
Since the ID is always unique the count will always be the same.

sarahfirstvoices: Good point about id. Would "pk" work here instead?

sarahfirstvoices (approved):

PR #539, opened 2023-09-21
FW-4949 N+1 query fixes

NolanVH:
### Description of Changes
This PR fixes several of the N+1 query issues including the categories list view and views that prefetch related media.

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/views/category_views.py
+++ b/firstvoices/backend/views/category_views.py
@@ -1,4 +1,4 @@
-import itertools
+from copy import deepcopy
 
 from django.db.models import Prefetch, Q
 from django.db.models.functions import Lower
@@ -162,8 +162,11 @@ class CategoryViewSet(SiteContentViewSetMixin, FVPermissionViewSetMixin, ModelVi
 
     def get_list_queryset(self):
         site = self.get_validated_site()
-        list_queryset = Category.objects.filter(site__slug=site[0].slug)
-        query = Q()
+        query = Q(site__slug=site[0].slug)
+
+        # Check for the nested flag
+        nested_flag_input = self.request.GET.get("nested", True)
+        nested_flag = get_valid_boolean(nested_flag_input)
 
         # Check if type flags are present
         contains_flags = [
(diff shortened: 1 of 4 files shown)


firstvoices/backend/views/category_views.py
```diff
@@ -172,46 +175,76 @@ def get_list_queryset(self):
             if len(flag)
         ]
 
-        # Check for the nested flag
-        nested_flag_input = self.request.GET.get("nested", True)
-        nested_flag = get_valid_boolean(nested_flag_input)
-
+        flags_query = Q()
         if len(contains_flags) > 0:
+            flags_query.connector = Q.OR
```
sahil97, line 180: just wondering, how this connector is used ? as I see that a connector is explicitly specified on line 184 as well for flags_query.

sahil97 (approved):
Looks good. +1 for the nested_query change.

firstvoices/backend/views/category_views.py
NolanVH: The connector is added there because behaviour seems to be different if it is left with the default AND connector. I was not getting the expected result without it. See https://stackoverflow.com/questions/25066894/django-q-add-weird-behavior-when-using-operator-or for more info.

firstvoices/backend/views/sites_views.py
```diff
@@ -79,7 +79,6 @@ def get_detail_queryset(self):
                 "menu",
                 "language",
                 "homepage",
-                "homepage",
```
sarahfirstvoices, line 82: Ha! Thanks for catching this.

I'm looking at the rest of this file and I realized that other_sites in the list method is missing prefetching. Could you add it there as well (it can match the prefetching from the Language query, just adjusted to work on a Site queryset.

sarahfirstvoices (approved):

firstvoices/backend/views/category_views.py
sahil97: Got it. Thanks

firstvoices/backend/views/sites_views.py
NolanVH: Good call. I have added prefetching for other_sites to this PR.

PR #559, opened 2023-09-28
FW-4873 Enhance Story update/patch API with page order editing

NolanVH:
### Description of Changes
This PR adds the ability to update page ordering in the Story update/patch APIs. The endpoints can be passed a "pages" field containing a list of page IDs in the desired order.

### Checklist

### Additional Notes
Note to @gmcauliffe and @Cara-Barter this will break any places on the frontend where the story update endpoint is used. A list of story page IDs will now need to be passed.

### DIFF
--- a/firstvoices/backend/serializers/story_serializers.py
+++ b/firstvoices/backend/serializers/story_serializers.py
@@ -146,21 +146,30 @@ class StoryDetailUpdateSerializer(StorySerializer):
                 updated_pages = validated_data["pages"]
                 existing_pages = instance.pages.all()
                 temp_story = Story.objects.create(site=instance.site)
-                for page in existing_pages:
-                    if page not in updated_pages:
-                        page.delete()
-                    else:
-                        page.story = temp_story
-                        page.save()
-                for index, page in enumerate(updated_pages):
-                    if page not in existing_pages:
-                        raise serializers.ValidationError(
-                            f"Page with ID {page.id} does not belong to the story."
-                        )
-                    else:
-                        page.ordering = index
-                        page.story = instance
-                        page.save()
+
+                # Get the intersection of the existing and updated pages
+                new_pages = [
+                    page for page in updated_pages if page in set(existing_pages)
+                ]
+
+                # Delete any existing pages that are not in the updated list to ensure no orphans
+                for page in [
+                    page for page in existing_pages if page not in updated_pages
+                ]:
+                    page.delete()
+
+                # Move the new pages to a temp story so that the order can be updated
+                # (the story and ordering is unique together)
+                for page in new_pages:
+                    page.story = temp_story
+                    page.save()
+
+                # Update the ordering and move the pages back to the original story
+                for index, page in enumerate(new_pages):
+                    page.ordering = index
+                    page.story = instance
+                    page.save()
+
                 temp_story.delete()
 
         return super().update(instance, validated_data)
--- a/firstvoices/backend/serializers/story_serializers.py
+++ b/firstvoices/backend/serializers/story_serializers.py
@@ -147,6 +147,13 @@ class StoryDetailUpdateSerializer(StorySerializer):
                 existing_pages = instance.pages.all()
                 temp_story = Story.objects.create(site=instance.site)
 
+                # Ensure that all updated pages belong to the story
+                for page in updated_pages:
+                    if page not in existing_pages:
+                        raise serializers.ValidationError(
+                            f"Page with ID {page.id} does not belong to the story."
+                        )
+
                 # Get the intersection of the existing and updated pages
                 new_pages = [
                     page for page in updated_pages if page in set(existing_pages)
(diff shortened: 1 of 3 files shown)


bronson-c (approved):
LGTM!

firstvoices/backend/serializers/story_serializers.py
```diff
@@ -125,6 +128,44 @@ class Meta(SiteContentLinkedTitleSerializer.Meta):
...
+                    if page not in existing_pages:
+                        raise serializers.ValidationError(
+                            f"Page with ID {page.id} does not belong to the story."
+                        )
+                    else:
+                        page.ordering = index
+                        page.story = instance
+                        page.save()
+                temp_story.delete()
+
```
sarahfirstvoices, line 165: I would double check this with @gmcauliffe but I think all pages must be created before trying to save the page order via this API.

So I don't think we need to do the temp_story bit, and I think we should validate that all the listed pages are *already* related to the current story instance, rather than setting the instance on them. I think we have 2 possible validation errors-- "Page with ID x does not exist" and "Page with ID x does not belong to this story"

firstvoices/backend/tests/test_apis/test_story_api.py
```diff
@@ -255,3 +256,67 @@ def test_pages_order(self):
...
+            "acknowledgements": [],
+            "hide_overlay": False,
+            "exclude_from_games": False,
+            "exclude_from_kids": False,
+            "related_audio": [],
+            "related_images": [],
+            "related_videos": [],
+        }
+
+        response = self.client.put(
```
sarahfirstvoices, line 306: Please test the patch for this-- it's what will be used 99% of the time for page order updates

sarahfirstvoices (changes requested):

firstvoices/backend/serializers/story_serializers.py
NolanVH: Thanks for the feedback! That was my understanding as well. 

The temp_story bit is there because the combination of a story and ordering is unique together so they cannot be updated in place. They are also required fields so we cannot set them to null and then update them after.

Any existing pages that aren't found in the updated list are deleted so that they aren't left as orphans. I have added a test to ensure this works correctly.

I have refactored this section to use the intersection of the existing and updated pages and I have added comments to make the process clear.

firstvoices/backend/serializers/story_serializers.py
```diff
@@ -125,6 +128,60 @@ class Meta(SiteContentLinkedTitleSerializer.Meta):
...
+            with transaction.atomic():
+                updated_pages = validated_data["pages"]
+                existing_pages = instance.pages.all()
+                temp_story = Story.objects.create(site=instance.site)
+
+                # Ensure that all updated pages belong to the story
+                for page in updated_pages:
+                    if page not in existing_pages:
+                        raise serializers.ValidationError(
+                            f"Page with ID {page.id} does not belong to the story."
```
sarahfirstvoices, line 154: The message here should be that the page does not exist, right?

firstvoices/backend/serializers/story_serializers.py
```diff
@@ -125,6 +128,60 @@ class Meta(SiteContentLinkedTitleSerializer.Meta):
...
+                        )
+
+                # Get the intersection of the existing and updated pages
+                new_pages = [
+                    page for page in updated_pages if page in set(existing_pages)
```
sarahfirstvoices, line 159: new_pages here will always result in the same list as updated_pages, due to the validation checking above, right?

firstvoices/backend/serializers/story_serializers.py
```diff
@@ -125,6 +128,60 @@ class Meta(SiteContentLinkedTitleSerializer.Meta):
...
+                ]
+
+                # Delete any existing pages that are not in the updated list to ensure no orphans
+                for page in [
```
sarahfirstvoices, line 163: This is going to iterate twice through the list-- better to just iterate once and move the condition inside

firstvoices/backend/serializers/story_serializers.py
sarahfirstvoices: That makes sense. I think the bulk_update method could be useful for this scenario, so we don't introduce the possibility of orphaning pages if there's an error while saving one of the pages. E.g., 

# validate the list of updated_pages
...

# set new page order on all page instances
for index, page in enumerate(updated_pages):
  page.ordering = index

# save all pages at once
StoryPage.objects.bulk_update(updated_pages, ['ordering'])

sarahfirstvoices: (works for me locally)

NolanVH: I originally tried `bulk_update` but was getting an integrity error due to the unique together constraint. I just tried it locally again and am still getting this error. It seems like `bulk_update` is not updating the pages in a single query in this case, at least for me.

Additionally, when using `bulk_update` the `pre_save` and `post_save` signals will not be sent which I believe are used to update `StoryPage` search indexing.

sarahfirstvoices (approved):

PR #588, opened 2023-10-11
Fw-4845, allow nulls for optional fields

sarahfirstvoices:
### Description of Changes
Allow nulls in all optional fields
- In json POST requests, nulls are just omitted. This is reflected in the tests.
- DRF handles this if fields are configured to be required=False and have a default value, so I updated the field configs and removed the NullableCharField we had previously used
- when there is no default for an optional field and the value is null, the default DRF behaviour is to skip setting that value on an update, which seems like a recipe for future bugs for us, so i added base tests and logged a separate ticket (FW-5065) to make sure each api is testing all the relevant values for POST, PUT, and PATCH. (Too much to deal with right now.) I did a pass on the dictionary_entry, song, and story APIs for now.

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/serializers/dictionary_serializers.py
+++ b/firstvoices/backend/serializers/dictionary_serializers.py
@@ -129,21 +129,27 @@ class DictionaryEntryDetailSerializer(
         many=True,
         required=False,
         source="acknowledgement_set",
+        default=[]
     )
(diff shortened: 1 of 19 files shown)


sahil97 (approved):
LGTM !

PR #591, opened 2023-10-12
FW-4681 Add contact-us API

NolanVH:
### Description of Changes
This PR adds a post endpoint to allow users to send contact us emails to a site.

If the `ENABLE_SMTP_BACKEND` environment variable is not set then emails will be sent and printed out to the console using the [Django console backend](https://docs.djangoproject.com/en/4.2/topics/email/#console-backend).

If the `ENABLE_SMTP_BACKEND` environment variable is set then emails will be sent using the [Django SMTP backend](https://docs.djangoproject.com/en/4.2/topics/email/#smtp-backend). Additional environment variables needed include `EMAIL_HOST` and `EMAIL_PORT` which specify the SMTP server to connect to, as well as `DEFAULT_FROM_EMAIL` which specifies the address emails will be sent from.

Emails are sent to all contacts in the `contact_email` and `contact_users` fields on the Site model. If no emails or users are found, a fallback email will be used which can be specified in the AppJson model. The fallback email can be set by adding a row to the AppJson model with the key `contact_us_default_emails` and the JSON field as a list of email strings.

A basic word exclusion list has been added. A row can be added to the AppJson model with the key `contact_us_excluded_words` and the JSON field as a list of word strings to be excluded. If any of these words are found in the contact-us form fields then the API will not send an email.

Finally, a contact-us "GET" endpoint has been added to allow language administrators to get the current list of email addresses to which contact-us emails will be sent.

### Checklist

### Additional Notes
N/A


firstvoices/backend/views/contact_us_views.py
```diff
@@ -0,0 +1,174 @@
...
+    def get_queryset(self):
+        return self.get_validated_site()
+
+    def list(self, request, *args, **kwargs):
+        """
+        Returns the list of receiver emails used by the contact us post endpoint.
+        """
+
+        site = self.get_validated_site().first()
+        perm = Site.get_perm("change")
```
sarahfirstvoices, line 76: This is a good choice for the permission, but please add a comment about it as well since it is different than our usual model permissions

firstvoices/backend/views/contact_us_views.py
```diff
@@ -0,0 +1,174 @@
...
+        return super().list(request, *args, **kwargs)
+
+    def create(self, request, *args, **kwargs):
+        """
+        Sends emails to the addresses listed on the site contact_us_email and contact_us_users fields. A list of
+        fallback email addresses can be set in the AppJson model with the key 'contact_us_default_emails'. If the
+        site does not have any contact emails or users, the fallback email will be used. Additionally, the site can
+        specify a list of excluded words in the AppJson model with the key 'contact_us_excluded_words'. If an email
+        contains any of these words in the email, name, or message it will not be sent.
+        """
```
sarahfirstvoices, line 88: 🙌🏻 Great docstring. Thank you for including all the details!

firstvoices/backend/views/contact_us_views.py
```diff
@@ -0,0 +1,174 @@
...
+
+        logger = logging.getLogger(__name__)
+        site = self.get_validated_site().first()
+        if request.user.is_anonymous:
+            return Response(
+                {"message": "You must be logged in to send an email."}, status=403
+            )
+        try:
+            to_email_list = site.contact_email
```
sarahfirstvoices, line 97: Oops, can you please rename contact_email to contact_emails since it is a list now?

firstvoices/backend/views/contact_us_views.py
```diff
@@ -0,0 +1,174 @@
...
+            # Validate that the message does not contain any excluded words
+            for word in word_list:
+                if re.search(word.lower(), (message + name + from_email).lower()):
+                    logger.error(
+                        f'Contact us message directed to site "{site.title}" contains excluded word. '
+                        "The message will not be sent."
+                    )
+                    raise ValidationError(
+                        "An excluded word has been found in the message. Please edit the "
+                        "message and try again."
```
sarahfirstvoices, line 173: Love this message. +1 for not including the offending word

firstvoices/backend/views/contact_us_views.py
```diff
@@ -0,0 +1,174 @@
...
+            word_list = excluded_words.first().json
+
+            # Validate that the excluded words are a list of strings
+            if not isinstance(word_list, list) or not all(
+                isinstance(word, str) for word in word_list
+            ):
+                logger.error(
+                    'The "contact_us_excluded_words" field in AppJson model must be a list of strings.'
+                )
+                raise ContactUsError()
```
sarahfirstvoices, line 162: I think we should use django's built in ImproperlyConfigured exception here.

firstvoices/backend/views/contact_us_views.py
```diff
@@ -0,0 +1,174 @@
...
+
+            # Validate that the message does not contain any excluded words
+            for word in word_list:
+                if re.search(word.lower(), (message + name + from_email).lower()):
```
sarahfirstvoices, line 166: Let's try this out... we may need to add some word boundaries to the search criteria so we don't match on the middle of words etc. E.g., if we banned the word "laugh" and someone used "manslaughter", or their email was based on a name like Lau Ghislaine or whatever.

firstvoices/backend/views/contact_us_views.py
```diff
@@ -0,0 +1,174 @@
...
+            )
+        try:
+            to_email_list = site.contact_email
+            to_email_list.extend(site.contact_users.values_list("email", flat=True))
+            name = request.data["name"]
+            from_email = request.data["email"]
+            message = request.data["message"]
+
+            # Validate that the email field from the form is a valid email address
+            validate_email(from_email)
```
sarahfirstvoices, line 104: Will this fail if the email address is blank or missing? Can you please add a test to verify that this field is required?

firstvoices/firstvoices/settings.py
```diff
@@ -314,3 +314,13 @@
 # Variables for the environment banners in the admin site
 ENVIRONMENT_NAME = os.getenv("SENTRY_ENVIRONMENT", "Local")
 ENVIRONMENT_COLOR = os.getenv("ENVIRONMENT_COLOR", "#9c9897")
+
+# Variables for the email backend (used in the contact us form)
+ENABLE_SMTP_BACKEND: bool = os.environ.get("ENABLE_SMTP_BACKEND") is not None
+DEFAULT_FROM_EMAIL = os.getenv("DEFAULT_FROM_EMAIL", "user@example.com")
```
sarahfirstvoices, line 320: Two thoughts:

1. I think this value should only have a fallback if smtp is not enabled. If we have an smtp server configured, it is reasonable to let this error out if the default_from_email value is missing.
2. Nothing ever overrides this from_email, so it might be confusing to call it a default. Maybe it could be called email_sender_address to match the other email_ variables?

firstvoices/backend/views/contact_us_views.py
```diff
@@ -0,0 +1,174 @@
...
+                    f"Message:\n{message}\n"
+                )
+                send_mail(
+                    subject=f"Contact Us Form Submission from {name} ({from_email})",
+                    message=final_message,
+                    recipient_list=to_email_list,
+                    from_email=settings.DEFAULT_FROM_EMAIL,
+                )
+            except (ConnectionRefusedError, SMTPException) as e:
+                logger.error(f"Contact us email failed to send. Error: {e}")
```
sarahfirstvoices, line 131: Please put the message into the exception and skip the error log. (Unhandled exceptions will be logged anyway.)

firstvoices/backend/views/contact_us_views.py
```diff
@@ -0,0 +1,174 @@
...
+                raise ContactUsError()
+
+            return Response(
+                {"message": "The email has been successfully sent."}, status=202
```
sarahfirstvoices, line 135: +1 for the 202 status, but please use the constant from rest_framework.status

sarahfirstvoices (changes requested):
Looking mostly good, Nolan! I added a lot of comments but none of them are major

NolanVH:
> Looking mostly good, Nolan! I added a lot of comments but none of them are major

I have made changes addressing all of your comments.
Thank you for the feedback @sarahfirstvoices!

README.md
```diff
@@ -99,6 +99,11 @@ License: Apache Software License 2.0
    `docker-compose up -d {service}` e.g.`docker-compose up -d elastic`.
    1. For ElasticSearch, to confirm the service is up and running, visit
       http://localhost:9200/ and verify the status.
+1. (Optional) By default, the App will send emails to console but an SMTP server can be used by setting the following environment variables:
+   - `ENABLE_SMTP_BACKEND`: Set to `True` to enable the SMTP backend.
+   - `DEFAULT_FROM_EMAIL`: The email address that emails will be sent from.
```
sarahfirstvoices, line 104: minor: this variable got renamed

firstvoices/backend/utils/contact_us_utils.py
```diff
@@ -0,0 +1,33 @@
...
+    contact_emails fields. The list of default emails are set in the AppJson model with the key
+    'contact_us_default_emails'.
+    """
+
+    logger = logging.getLogger(__name__)
+    if AppJson.objects.filter(key="contact_us_default_emails").count() == 0:
+        logger.error(
+            'No default email set in AppJson model with key "contact_us_default_email".'
+        )
+        raise ContactUsError()
```
sarahfirstvoices, line 19: Sorry, I must have missed this one-- this should probably be an ImproperlyConfigured exception with the message in the exception instead of in a log. Same for the one below

sarahfirstvoices (approved):
Sorry for a second round of tiny feedback! No need to re-review if you make the changes. Looks good to me :)

PR #602, opened 2023-10-17
FW-4671 media search APIs

sahil97:
### Description of Changes
- Added code to search for media files through the search API
- Added code to remove media items from index when the site is deleted
- refactored hydration logic into a separate file
- updated tests to reflect changes

### Checklist

### Additional Notes
N/A


firstvoices/backend/search/utils/query_builder_utils.py
```diff
@@ -54,13 +56,15 @@ def get_cleaned_search_term(q):
 
 # sub-queries utils
 def get_types_query(types):
```
sahil97, line 58: Refactored to make it more generic and include any possible future cases.

firstvoices/backend/tests/test_search/test_object_utils.py
```diff
@@ -0,0 +1,319 @@
...
+        document_type = "dictionary_entry"
+        document_id = "0101"
+
+        result = get_object_from_index(test_index, document_type, document_id)
+
+        assert result is None
+
+    @pytest.mark.skip(reason="Cannot emulate the connection error.")
+    def test_connection_error(self):
+        # Skipping as could not emulate the connection error. ES client always raises the connection timeout
```
sahil97, line 84: Open to any suggestions on how to test these error cases.

firstvoices/backend/search/utils/hydration_utils.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+def separate_object_ids(search_results):
+    # Separate the mixed search results to fetch objects from db
+    search_results_dict = {
+        ELASTICSEARCH_DICTIONARY_ENTRY_INDEX: [],
+        ELASTICSEARCH_SONG_INDEX: [],
+        ELASTICSEARCH_STORY_INDEX: [],
+        ELASTICSEARCH_MEDIA_INDEX: {
+            "audio": [],
```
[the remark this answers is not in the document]
NolanVH: I think either option would be okay here but it wouldn't hurt to define them as constants just in case we decide to refactor or change the names down the road.

firstvoices/backend/tests/test_search/test_object_utils.py
NolanVH: Had a quick chat about how we might be able to test these cases.

NolanVH: Had a quick chat on how we might be able to test these.

NolanVH (commented):
These changes are looking great! Tested working locally.

The API view doc string decorators could probably get updated to give some information on how media can be searched. The Postman workspace could also use a small update.

Will re-review after the tests have been updated as discussed.

firstvoices/backend/search/signals/media_signals.py
```diff
@@ -37,6 +37,4 @@ def request_update_media_index(sender, instance, **kwargs):
 @receiver(post_delete, sender=Image)
 @receiver(post_delete, sender=Video)
 def request_delete_from_index(sender, instance, **kwargs):
-    transaction.on_commit(
-        delete_from_index.apply_async((instance.id,), link_error=link_error_handler.s())
-    )
+    delete_from_index.apply_async((instance.id,), link_error=link_error_handler.s())
```
sarahfirstvoices, line 40: Was waiting on the transaction a problem here?

firstvoices/backend/search/utils/hydration_utils.py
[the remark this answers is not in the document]
sarahfirstvoices: Constants sound good to me. I don't think they need to be an enum, and I think they can be specific to the search code.

Alternative idea, just to consider: would it be simpler to just put them in separate indices so we don't have to track the type at all? The other model types are all separate. Is there a benefit to merging the media models into one index?

sarahfirstvoices (approved):

firstvoices/backend/search/signals/media_signals.py
sahil97: Yes, so we lose the instance after the transaction gets completed, and thus the instance.id cannot be passed down.
I discussed this with Bronson for confirmation, and the race condition only existed during the save action. Thus, the transaction.on_commit could be safely removed from here.

firstvoices/backend/search/utils/hydration_utils.py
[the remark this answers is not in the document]
sahil97: Yes, so I had the idea of using separate indices as well. My reasoning for putting them in one index was that the document structure is same for all media files with just the difference in `type`.

Separate indices:
- Upside: is that we can get some performance improvement, I'm not sure by how much, since my understanding is that ES only scans the inverted index during a `filter` operation and does not scan all the documents, so increasing the document count should not have a major impact. 
- Downside: Extra methods and overhead to manage 2 more indices. There will not be a lot of repeated code, which is good, since the document structure and the CUD operations remains the same for all 3 indices. But, I guess there will be more conditional clauses in CUD methods related to this where the the index name is checked instead of type.

One index:
- Upside: Less code, same document structure.
- Downside: If `filter` operation does affect the performance majorly, we can move to separate indices. I'm positive it may have some impact, since it is an additional operation after choosing the index, but after researching a bit, still not sure if it's a major time taker.

sahil97:
> These changes are looking great! Tested working locally.
> 
> The API view doc string decorators could probably get updated to give some information on how media can be searched. The Postman workspace could also use a small update.
> 
> Will re-review after the tests have been updated as discussed.

Thanks for the suggestions. The API doc strings and postman workspace was updated in a previous PR related to media search API.
I've added the tests similar to your PR as well. Please re-review when you get some time. Thanks.

NolanVH (approved):
All changes look good to me!

firstvoices/backend/search/signals/media_signals.py
bronson-c: Yes, none of the delete from index signal have/require on_commit to function properly

bronson-c (approved):
LGTM! Great work!

PR #605, opened 2023-10-18
Fw-4546, approve/reject/ignore APIs for join requests

sarahfirstvoices:
### Description of Changes
- added these APIs as additional POST endpoints nested under /join-requests
- bonus change: tweaked the permissions for join requests

### Checklist
- [x] API docs updated

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/permissions/predicates/base.py
+++ b/firstvoices/backend/permissions/predicates/base.py
@@ -1,4 +1,4 @@
-from rules import predicate
+from rules import predicate, Predicate
 
 from backend.models.constants import AppRole, Role, Visibility
 from backend.permissions.utils import get_app_role, get_site_role
@@ -99,9 +99,15 @@ def is_superadmin(user, obj):
 
 
 #
+# site and app role combos
+is_at_least_language_admin = Predicate(
+    has_language_admin_membership | is_at_least_staff_admin, name="is_at_least_language_admin"
+)
+
+#
 # access-based predicates
 #
-has_public_access_to_obj = predicate(
+has_public_access_to_obj = Predicate(
     is_public_obj & has_public_site, name="has_public_access_to_obj"
 )
 has_member_access_to_obj = (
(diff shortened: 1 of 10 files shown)


firstvoices/backend/permissions/predicates/base.py
```diff
@@ -1,4 +1,4 @@
-from rules import predicate
+from rules import predicate, Predicate
```
sarahfirstvoices, line 1: This is correcting an error that wasn't causing problems before-- we were creating some predicates using the function decorator (lowercase p) instead of the class (uppercase p). Main effect of that is that using the decorator as a function doesn't name the resulting predicate, so it couldn't get matched up to a filter with the same name. Now these can all be used as view permissions on models without triggering a configuration warning about a missing permission filter

firstvoices/backend/serializers/join_request_serializers.py
```diff
@@ -13,7 +16,7 @@
 
 
 class JoinRequestDetailSerializer(BaseSiteContentSerializer):
-    url = SiteHyperlinkedIdentityField(view_name="api:join-request-detail")
+    url = SiteHyperlinkedIdentityField(view_name="api:joinrequest-detail")
```
sarahfirstvoices, line 19: I changed this only because if django ever tried to auto-generate a url for a join request, it would use the non-hyphenated version (just lowercases the model name)

firstvoices/backend/urls.py
```diff
@@ -57,14 +57,14 @@
     r"ignored-characters", IgnoredCharactersViewSet, basename="ignoredcharacter"
 )
 sites_router.register(r"images", ImageViewSet, basename="image")
+sites_router.register(r"join-requests", JoinRequestViewSet, basename="joinrequest")
```
bronson-c, line 60: The postman endpoints should be updated to reflect the removal of the hyphen

bronson-c (approved):
LGTM! Great catch with the predicates and the unit tests are great. Only thing I spotted was changing our Postman requests to reflect the endpoint changes 👍

NolanVH (approved):
All changes look good to me!

firstvoices/backend/urls.py
sarahfirstvoices: I didn't change the url, only the internal "basename" used to generate the viewnames (like `api:joinrequest-detail` for reversing urls). So Postman should be good still :)

PR #619, opened 2023-10-23
Added a bug fix for rebuild index for media files

sahil97:
### Description of Changes
Bug fix for missing attributes when indexing media.

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/management/commands/utils/iterators.py
+++ b/firstvoices/backend/management/commands/utils/iterators.py
@@ -35,19 +35,37 @@ def media_doc_generator(instance, media_type):
 def audio_iterator():
     queryset = Audio.objects.all()
     for instance in queryset:
-        yield media_doc_generator(instance, "audio").to_dict(True)
+        try:
+            yield media_doc_generator(instance, "audio").to_dict(True)
+        except AttributeError:
+            print(
+                f"Skipping document due to missing properties. object: audio, id: {instance.id}."
+            )
+            continue
 
 
 def image_iterator():
     queryset = Image.objects.all()
     for instance in queryset:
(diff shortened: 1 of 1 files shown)


firstvoices/backend/management/commands/utils/iterators.py
```diff
@@ -35,19 +35,37 @@ def media_doc_generator(instance, media_type):
 def audio_iterator():
     queryset = Audio.objects.all()
     for instance in queryset:
-        yield media_doc_generator(instance, "audio").to_dict(True)
+        try:
+            yield media_doc_generator(instance, "audio").to_dict(True)
+        except AttributeError:
+            print(
```
NolanVH, line 41: These should probably be logs instead of printouts. Same with the two functions below.

NolanVH (approved):
Aside from the one minor comment, all changes look good to me!

firstvoices/backend/management/commands/utils/iterators.py
sahil97: Good idea. The reason these are print statements are because this command is always executed from the shell manually so it never goes to a logger or sentry. I used self.stdout.write in the main method to log the "success messages" but to use that I would have to pass the "self" object through multiple methods to reach this point.
I did a bit of research, and print is almost the same as stdout.write (print converts whatever you give it to a string, and then calls the stdout.write internally) so I left it at print.
My main focus to get this merged is to find the problematic files which are missing attributes.

sahil97: I did some more research, while the main was building to reflect the changes. Seems like logger is the way to go. I will update it and raise another PR to add those changes.
Thanks for the suggestion.

PR #627, opened 2023-10-25
FW-4945 Add usage throttling

NolanVH:
### Description of Changes
This PR adds usage throttling. 

As a start, there is both a sustained rate and a burst rate which can be set with the environment variables. They are set as strings in the form of "#/time" where # is the number of requests and time is sec/min/hour/day (eg. "60/min" or "1000/day"). The environment variables are `SUSTAINED_THROTTLE_RATE` and `BURST_THROTTLE_RATE`. If no environment variables are set the sustained rate defaults to 1000 requests per day and the burst rate defaults to 60 requests per minute. 

The variables are set to very high numbers by default if debug is enabled (for local development).

Per user role/API throttling can be added after we launch the new version and have detailed usage statistics.

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/views/base_views.py
+++ b/firstvoices/backend/views/base_views.py
@@ -9,7 +9,15 @@ from backend.permissions import utils
 from backend.views.utils import BurstRateThrottle, SustainedRateThrottle
 
 
-class FVPermissionViewSetMixin:
+class ThrottlingMixin:
+    """
+    A mixin to provide request usage throttling for viewsets.
+    """
+
+    throttle_classes = [BurstRateThrottle, SustainedRateThrottle]
+
+
+class FVPermissionViewSetMixin(ThrottlingMixin):
     """
     Forked from ``rules.contrib.rest_framework.AutoPermissionViewSetMixin`` to provide extension points.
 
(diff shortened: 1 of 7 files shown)


firstvoices/backend/views/base_views.py
```diff
@@ -129,6 +131,8 @@ class SiteContentViewSetMixin:
     Provides common methods for handling site content, usually for data models that use the ``BaseSiteContentModel``.
     """
 
+    throttle_classes = [BurstRateThrottle, SustainedRateThrottle]
+
```
sarahfirstvoices, line 135: Hmm. We actually have a couple of APIs that don't use either of these mixins (sites list, I think?). I don't want to have to add a new mixin to every API... pondering the best way to handle this.

sarahfirstvoices: Maybe make a ThrottlingMixin and add that to the FVPermissions Mixin as well as our non-FVPermissions APIs? I don't think we have any APIs that use SiteContentMixin without FVPermissionsMixin, so could probably skip adding it in the SiteContent one.

sarahfirstvoices (approved):

firstvoices/backend/views/base_views.py
bronson-c: +1 to this solution

bronson-c (approved):

PR #662, opened 2023-11-08
FW-5138 Add order by search parameter

NolanVH:
### Description of Changes
This PR adds the ability to order the search results by the following:
- `created`: orders results by the created timestamp
- `modified`: orders results by the last modified timestamp
- `title`: orders results by title based on the alphabet custom order

Results are sorted in ascending order by default and `_desc` can optional get added to the parameter which will sort results in descending order (eg: `sort=created_desc`).

The created and last_modified values have been added to the search index to support these parameters. Additionally, the `created` and `last_modified` values have been added to the serializers used by the search view.

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/search/utils/query_builder_utils.py
+++ b/firstvoices/backend/search/utils/query_builder_utils.py
@@ -281,7 +281,7 @@ def get_valid_visibility(input_visibility_str):
 
 
 def get_valid_sort(input_sort_by_str):
-    input_string = input_sort_by_str.strip().split("_")
+    input_string = input_sort_by_str.lower().strip().split("_")
 
     descending = len(input_string) > 1 and input_string[1] == "desc"
 
@@ -292,4 +292,4 @@ def get_valid_sort(input_sort_by_str):
     ):
         return input_string[0], descending
     else:  # if invalid string is passed
-        return None, descending
+        return None, None
--- a/firstvoices/backend/search/utils/query_builder_utils.py
+++ b/firstvoices/backend/search/utils/query_builder_utils.py
@@ -278,3 +278,19 @@ def get_valid_visibility(input_visibility_str):
     if len(selected_values) == 0:
         return None
     return selected_values
+
+
+def get_valid_order_by(input_order_by_str):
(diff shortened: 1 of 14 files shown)


sarahfirstvoices (changes requested):

bronson-c (approved):
LGTM! 👍

firstvoices/backend/search/utils/query_builder_utils.py
```diff
@@ -278,3 +278,18 @@ def get_valid_visibility(input_visibility_str):
...
+    descending = len(input_string) > 1 and input_string[1] == "desc"
+
+    if len(input_string) > 0 and (
+        input_string[0] == "created"
+        or input_string[0] == "modified"
+        or input_string[0] == "title"
+    ):
+        return input_string[0], descending
+    else:  # if invalid string is passed
+        return None, descending
```
sarahfirstvoices, line 295: This is fine, but I don't think we really need to return the descending value here

sarahfirstvoices: I take it back about this being fine-- I think we should return None, None here to eliminate the possibility of a bug that would sort by reverse search score

firstvoices/backend/views/search/base_search_views.py
```diff
@@ -379,10 +425,44 @@ def list(self, request, **kwargs):
             from_=pagination_params["start"], size=pagination_params["page_size"]
         )
 
-        # Sort by score, then by custom sort order
-        search_query = search_query.sort(
-            "_score", {"custom_order": {"unmapped_type": "keyword"}}, "title.raw"
+        sort_direction = (
+            "desc"
+            if search_params["descending"] and search_params["sort"] is not None
```
sarahfirstvoices, line 430: If you always return None for descending when the sort param is missing or invalid, you can skip this type of check on the sort param value

sarahfirstvoices (approved):
Thanks for making all these changes! I'm approving it, but please do update that one return value

PR #664, opened 2023-11-10
Fw 5154 dictionary recalc move db operations into async tasks

sahil97:
### Description of Changes
- Moved the db ops to async tasks
- Added a check before starting the next task to verify if there is a task already running.
- Added a separate ticket to JIRA to refactor and reenable the skipped tests.  [FW-5207](https://firstvoices.atlassian.net/browse/FW-5207)

### Checklist

### Additional Notes
N/A

[FW-5207]: https://firstvoices.atlassian.net/browse/FW-5207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/firstvoices/backend/serializers/async_results_serializers.py
+++ b/firstvoices/backend/serializers/async_results_serializers.py
@@ -22,16 +22,26 @@ class CustomOrderRecalculationResultSerializer(serializers.ModelSerializer):
     @staticmethod
     @extend_schema_field(OpenApiTypes.STR)
     def get_latest_recalculation_result(obj):
+        if (
+            hasattr(obj, "latest_recalculation_result")
+            and obj.latest_recalculation_result
+        ):
+            unknown_character_count = obj.latest_recalculation_result.get(
+                "unknown_character_count", 0
+            )
+            updated_entries = obj.latest_recalculation_result["updated_entries"]
+        else:
+            unknown_character_count = 0
+            updated_entries = []
+
         ordered_result = OrderedDict(
             {
-                "unknown_character_count": obj.latest_recalculation_result[
-                    "unknown_character_count"
-                ],
+                "unknown_character_count": unknown_character_count,
                 "updated_entries": [],
             }
         )
 
-        for entry in obj.latest_recalculation_result["updated_entries"]:
+        for entry in updated_entries:
             ordered_entry = OrderedDict(
                 {
                     "title": entry["title"],
--- a/firstvoices/backend/tests/test_apis/test_dictionary_cleanup_api.py
+++ b/firstvoices/backend/tests/test_apis/test_dictionary_cleanup_api.py
@@ -108,6 +108,9 @@ class TestDictionaryCleanupPreview(BaseApiTest):
         assert response.status_code == 202
         assert response_data == self.SUCCESS_MESSAGE
 
+    @pytest.mark.skip(
+        reason="This test now checks async code which can be done in a separate integration test suite."
+    )
     @pytest.mark.django_db(transaction=True, serialized_rollback=True)
     def test_recalculate_e2e(
         self,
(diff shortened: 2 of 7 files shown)


firstvoices/backend/tests/test_apis/test_dictionary_cleanup_api.py
```diff
@@ -108,6 +108,9 @@ def test_recalculate_post(
         assert response.status_code == 202
         assert response_data == self.SUCCESS_MESSAGE
 
+    @pytest.mark.skip(
```
sahil97, line 111: Skipping these tests for the following reason.
Originally, when the view operated synchronously, the creation of the database entry tested by the retrieve API was always created on time. However, moving all operations into asynchronous tasks introduces uncertainty regarding precisely when the database entry is added, even though it is the first operation within the task.
We also cannot add wait/sleep here, as that can introduce flakiness in the tests. Though, just by adding a sleep(1) worked always on my local, but I believe its better not to introduce that and include this test in the integrations tests rather than here since its is now testing a async behaviour.

firstvoices/backend/serializers/async_results_serializers.py
```diff
@@ -22,16 +22,26 @@ def get_current_task_status(obj):
     @staticmethod
     @extend_schema_field(OpenApiTypes.STR)
     def get_latest_recalculation_result(obj):
+        if (
```
bronson-c, line 25: I like this change/adding defaults!

bronson-c (approved):
LGTM, thanks for fixing the non-async parts of this process!

firstvoices/backend/views/custom_order_recalculate_views.py
```diff
@@ -67,21 +68,26 @@ def create(self, request, *args, **kwargs):
 
         # Call the recalculation task
         try:
-            recalculation_results = recalculate_custom_order.apply_async((site_slug,))
-
-            # Delete any previous results
-            CustomOrderRecalculationResult.objects.filter(
+            # Check if preview task for the same site is ongoing
```
caforbes, line 71: super minor: this comment is about the preview, but this code is for the real recalculation.

caforbes (approved):
Seems good to me as long as we have a plan for where to re-add the skipped tests! :)

firstvoices/backend/tests/test_apis/test_dictionary_cleanup_api.py
sarahfirstvoices: How about setting celery to run synchronously for the tests? We do that for a few others-- e.g., media thumbnail tests.

firstvoices/backend/views/custom_order_recalculate_views.py
```diff
@@ -67,21 +68,26 @@ def create(self, request, *args, **kwargs):
...
+                    status = AsyncResult(task.task_id).status
+                    if status == "PENDING":
+                        running_tasks += 1
+
+            if running_tasks > 0:
+                return Response(
+                    {
+                        "message": "Recalculation is already running. Please use retrieve API to verify status."
+                    },
+                    status=202,
```
sarahfirstvoices, line 87: I think this is a good workflow, but this is not the right status code for this type of response. If we want to tell someone to GET another URL instead of processing this POST, we should return status 303 (see other) and put the destination URL in the Location header. https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections

firstvoices/backend/views/custom_order_recalculate_views.py
```diff
@@ -139,23 +145,26 @@ def create(self, request, *args, **kwargs):
...
+                    status = AsyncResult(task.task_id).status
+                    if status == "PENDING":
+                        running_tasks += 1
+
+            if running_tasks > 0:
+                return Response(
+                    {
+                        "message": "Cleanup preview is already running. Please use retrieve API to verify status."
+                    },
+                    status=202,
```
sarahfirstvoices, line 164: Same comment as above-- this should be a 303 response with a Location header.

sarahfirstvoices (changes requested):

sarahfirstvoices (approved):

PR #671, opened 2023-11-15
Fw 5101 fix remaining n 1 query api issues

sahil97:
### Description of Changes
- Added caching to 3 properties of the Alphabet model.
- Added a select_related for variant_characters property to resolve N+1 query issue.
- Moved the prefetch_related clause to select_related for ImageViewSet and VideoViewSet to resolve N+1 query issue.
- Added a separate ticket to refactor the base_views to further reduce duplicate and similar queries. Ticket: [FW-5208](https://firstvoices.atlassian.net/browse/FW-5208)

### Checklist

### Additional Notes
N/A

[FW-5208]: https://firstvoices.atlassian.net/browse/FW-5208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/firstvoices/backend/models/characters.py
+++ b/firstvoices/backend/models/characters.py
@@ -5,6 +5,7 @@ import yaml
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.db.utils import IntegrityError
+from django.utils.functional import cached_property
 from django.utils.translation import gettext as _
 
 from backend.permissions import predicates
@@ -239,21 +240,23 @@ class Alphabet(BaseSiteContentModel):
     # JSON representation of a g2p mapping from confusable characters to canonical characters
     input_to_canonical_map = models.JSONField(default=list)
 
-    @property
+    @cached_property
     def base_characters(self):
         """
         Characters for the site in sort order.
         """
         return Character.objects.filter(site=self.site).order_by("sort_order")
 
-    @property
+    @cached_property
     def variant_characters(self):
         """
         All variant characters for the site.
         """
-        return CharacterVariant.objects.filter(site=self.site)
+        return CharacterVariant.objects.filter(site=self.site).select_related(
+            "base_character"
+        )
 
-    @property
+    @cached_property
     def ignorable_characters(self):
         """
         Ignorable characters for the site.
--- a/firstvoices/backend/views/image_views.py
+++ b/firstvoices/backend/views/image_views.py
@@ -81,7 +81,7 @@ class ImageViewSet(
         site = self.get_validated_site()
         return (
             Image.objects.filter(site__slug=site[0].slug)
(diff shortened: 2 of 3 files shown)


firstvoices/backend/models/characters.py
```diff
@@ -239,21 +240,23 @@ class Meta:
     # JSON representation of a g2p mapping from confusable characters to canonical characters
     input_to_canonical_map = models.JSONField(default=list)
 
-    @property
+    @cached_property
```
sahil97, line 243: So these 3 properties were being used in multiple places, and since all these properties are DB calls, we were getting multiple similar queries for the 3 related models - `Character, CharacterVariant and IgnoredCharacter`.
Using a `@cached_property` reduces those duplicate queries as the properties only get evaluated once for an instance.

The only catch is that we should not be modifying any of these 3 related models within the same request/response cycle 
or
Any other place where we are using an `Alphabet` instance and we modify any of the 3 models, and then want to use the updated results. In that case, we will have to manually refresh the property.

I did glance through the code where we are using `Alphabet` model, and also verified with Bronson if it is safe to cache these properties. All the tests are also passing. 
But please inform if there are any edge cases where we can run into problem.
Thanks :)

NolanVH (approved):
LGTM!

firstvoices/backend/views/image_views.py
```diff
@@ -81,7 +81,7 @@ def get_queryset(self):
         site = self.get_validated_site()
         return (
             Image.objects.filter(site__slug=site[0].slug)
-            .prefetch_related("site", *get_select_related_media_fields(None))
+            .select_related(*get_select_related_media_fields(None))
```
sarahfirstvoices, line 84: If I recall, Nolan switched this because although it added additional queries it reduced the overall response time. Could you verify that this is an improvement in response time as well as the query count?

firstvoices/backend/views/video_views.py
```diff
@@ -81,7 +81,7 @@ def get_queryset(self):
         site = self.get_validated_site()
         return (
             Video.objects.filter(site__slug=site[0].slug)
-            .prefetch_related("site", *get_select_related_media_fields(None))
+            .select_related(*get_select_related_media_fields(None))
```
sarahfirstvoices, line 84: Same question about prefetch vs select

firstvoices/backend/views/image_views.py
sahil97: Yes, I'm not sure about other views, but I have verified that there is no time difference in using select or prefetch related in these couple of viewsets.

PR #678, opened 2023-11-17
FW-4347 enhance site data export API

NolanVH:
### Description of Changes
This PR adds the following changes to the site data API at `/api/1.0/sites/<site  slug>/data`:
- Removed nesting of `data` and `config` inside of `siteDataExport` field
- Add categories list
- Updated audio and image data on `data` entries to now hold actual data instead of placeholder data
- Add pagination to `data` field
- Fixed N+1 query issues and optimized endpoint

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/serializers/site_data_serializers.py
+++ b/firstvoices/backend/serializers/site_data_serializers.py
@@ -1,6 +1,8 @@
+from collections import OrderedDict
 from datetime import datetime
 
-from rest_framework import serializers
+from django.core.paginator import Paginator
+from rest_framework import pagination, serializers
 
 from backend.models import Category, Site
 from backend.models.dictionary import DictionaryEntry, TypeOfDictionaryEntry
@@ -48,10 +50,23 @@ class SiteDataSerializer(SiteContentLinkedTitleSerializer):
             else None
         )
         request = self.context.get("request")
-        dictionary_entries = DictionaryEntryDataSerializer(
+        serializer = DictionaryEntryDataSerializer(
             queryset, many=True, context={"request": request}
-        ).data
+        )
+
+        paginator = DictionaryEntryPaginator()
+        paginated_data = paginator.paginate_queryset(
+            queryset=serializer.data, request=request
+        )
+        dictionary_entries = paginator.get_paginated_response(paginated_data)
         return dictionary_entries
+
+    def to_representation(self, instance):
+        output = super().to_representation(instance)
+        data = output.pop("data", None)
+        for key, value in data.items():
+            output[key] = value
+        return output
 
     class Meta:
         model = Site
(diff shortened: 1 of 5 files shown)


firstvoices/backend/serializers/site_data_serializers.py
```diff
@@ -38,86 +54,100 @@ def get_site_data_export(self, site):
...
-    def get_source(self, dictionaryentry):
+    @staticmethod
+    def get_source(dictionaryentry):
         return dict_entry_type_mtd_conversion(dictionaryentry.type)
 
-    def get_definition(self, dictionaryentry):
-        if dictionaryentry.translation_set.first() is not None:
-            return dictionaryentry.translation_set.first().text
+    @staticmethod
+    def get_definition(dictionaryentry):
```
bronson-c, line 112: Question out of curiosity, what if the dictionary entry has more that one translation? Does only the first need to be provided? If I recall correctly some of the migrated entries also had translations that were split into two for being too long, but that may only apply to longer models like songs/stories

firstvoices/backend/views/data_views.py
```diff
@@ -44,25 +52,65 @@ class SitesDataViewSet(
...
+                .prefetch_related(
+                    "speakers",
+                ),
+            ),
+            Prefetch(
+                "related_images",
+                queryset=Image.objects.all().select_related(
+                    "original",
+                ),
+            ),
```
sarahfirstvoices, line 79: Do we also need to prefetch videos?

firstvoices/backend/views/data_views.py
```diff
@@ -44,25 +52,65 @@ class SitesDataViewSet(
...
             )
         )
-        return sites
+
+        site_config_and_categories_json = SiteDataSerializer(
+            site[0], context={"request": request}
+        ).data
+
+        for key, value in site_config_and_categories_json.items():
+            data[key] = value
```
sarahfirstvoices, line 109: This is totally fine, but just a heads up that this kind of thing is a good use for the python `...` magic, like `data = { ...other_dictionary }`

sarahfirstvoices (approved):
Looks good! Thanks for making those big reorganizations. I added a couple optional suggestions but nothing that needs a re-review

firstvoices/backend/views/data_views.py
NolanVH: The API doesn't currently return video data as that was not in the ticket requirements. It could be added if needed though.

firstvoices/backend/serializers/site_data_serializers.py
NolanVH: The MTD format appears to only specify a single definition right now but this is a good point and could be a good improvement for a future ticket.

PR #697, opened 2023-11-28
FW-4333 Add Immersion Label Data Model

bronson-c:
### Description of Changes
- Adds the immersion label model with a unique key per site and dictionary entry foreign key
- Adds immersion labels to admin site (can be removed once API is in place)
- Adds uniqueness test for immersion label model
### Checklist

### Additional Notes
Requirements for the data model can be found here: https://firstvoices.atlassian.net/wiki/spaces/FIR/pages/419071188/Immersion+Label+Tech+Design


firstvoices/backend/models/immersion_labels.py
```diff
@@ -0,0 +1,39 @@
...
+        }
+        indexes = [
+            models.Index(fields=["site", "key"], name="immersion_label_key_idx"),
+        ]
+
+    key = models.CharField(max_length=DEFAULT_TITLE_LENGTH)
+
+    dictionary_entry = models.ForeignKey(
+        "DictionaryEntry",
+        on_delete=models.CASCADE,
```
sahil97, line 37: I'm not sure if we want to have CASCADE here. In case the relevant dictionary_entry gets deleted, this row would be deleted as well, right ?
Should we use SET_NULL instead so as to save the key and the user can later change to some other dictionary entry ?

bronson-c: This is a better solution, I'll change it. Thanks!

sahil97 (approved):

PR #709, opened 2023-12-04
FW-5253 Add read/write API for site features

bronson-c:
### Description of Changes
- Moves site features to their own read/write API: Standard CRUD + patch functions on site features
- Removes old site feature serialization for site models

### Checklist

### Additional Notes
N/A


sahil97 (approved):
LGTM !

firstvoices/backend/views/site_feature_views.py
```diff
@@ -0,0 +1,111 @@
...
+    ),
+)
+class SiteFeatureViewSet(
+    SiteContentViewSetMixin, FVPermissionViewSetMixin, ModelViewSet
+):
+    """
+    API endpoint that allows site feature flags to be viewed or edited.
+    """
+
+    lookup_field = "key"
```
NolanVH, line 101: If the lookup field is the key and the key is allowed to have any characters then the detail view will have the strange characters in the slug (eg: if the key is "test key" then the slug will be `.../<site-slug>/features/test key`, which we probably don't want).

We may want to either restrict the format of the key field or add another field to hold the slug.

firstvoices/backend/serializers/site_serializers.py
```diff
@@ -17,25 +17,16 @@
...
     """
 
     logo = ImageSerializer()
-    features = FeatureFlagSerializer(
-        read_only=True, source="sitefeature_set", many=True
-    )
 
     class Meta(LinkedSiteSerializer.Meta):
-        fields = LinkedSiteSerializer.Meta.fields + ("logo", "features")
+        fields = LinkedSiteSerializer.Meta.fields + ("logo",)
```
sarahfirstvoices, line 29: Let's not remove this please-- it is a specially formatted list of only the enabled features for a site, currently used by the frontend for things like immersion mode

sarahfirstvoices: Maybe we could rename it enabledFeatures or something, but we need to let Guy know.

firstvoices/backend/tests/test_apis/test_site_features_api.py
```diff
@@ -0,0 +1,159 @@
...
+    def get_expected_response(self, instance, site):
+        return self.get_expected_detail_response(instance, site)
+
+    def get_valid_data(self, site=None):
+        return {
+            "key": self.TEST_KEY,
+            "isEnabled": True,
+        }
+
+    def get_valid_data_with_nulls(self, site=None):
```
sarahfirstvoices, line 56: Rather than having this fake test, please just override and skip the null value tests

firstvoices/backend/tests/test_apis/test_site_features_api.py
```diff
@@ -0,0 +1,159 @@
...
+
+        response = self.client.put(
+            self.get_detail_endpoint(key=feature.key, site_slug=site.slug),
+            format="json",
+            data=data,
+        )
+
+        assert response.status_code == 200
+        response_data = response.json()
+        assert response_data["key"] == feature.key
```
sarahfirstvoices, line 137: Great test case

sarahfirstvoices (changes requested):
Marking this "request changes" to get the list of enabled features back in the sites api. Thanks!

NolanVH (approved):
LGTM! Thanks for the updates.

firstvoices/backend/serializers/site_serializers.py
bronson-c: I have re-added the features field to the site serializer as enabledFeatures and I let Guy know about the change!

sarahfirstvoices (approved):

PR #748, opened 2024-01-06
FW-5402 Add wordsy endpoint

sahil97:
### Description of Changes
- Added endpoint to generate config for _wordsy_ game at `/{site_slug}/wordsy`.
- Uses cache for querysets to preserve the config for the given calendar date.

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/tests/test_apis/test_games_api.py
+++ b/firstvoices/backend/tests/test_apis/test_games_api.py
@@ -14,7 +14,7 @@ from backend.tests.test_apis.base_api_test import BaseSiteContentApiTest
 
 class TestWordsyEndpoint(BaseSiteContentApiTest):
     """
-    Tests for games/wordsy API
+    Tests for wordsy endpoint
     """
 
     API_LIST_VIEW = "api:wordsy-list"
@@ -44,7 +44,7 @@ class TestWordsyEndpoint(BaseSiteContentApiTest):
 
     @pytest.mark.django_db
     def test_all_invalid_words(self):
-        invalid_words_list = ["abc ab", "abcabc", "xyzav"]
+        invalid_words_list = ["abc ab", "abcabc", "xyzav", "ab ab"]
         # adding multiple dictionary entries
         for title in invalid_words_list:
             DictionaryEntryFactory(
@@ -64,7 +64,7 @@ class TestWordsyEndpoint(BaseSiteContentApiTest):
     @pytest.mark.django_db
     def test_mixed_words(self):
         # test to verify only valid words show up
-        invalid_words_list = ["abc ab", "abcabc", "xyzav"]
+        invalid_words_list = ["ab ab", "abc ab", "abcabc", "xyzav"]
         valid_words_list = ["ababa", "abcab", "bcbca", "cabca"]
 
         # adding multiple dictionary entries
--- a/firstvoices/backend/tests/test_apis/test_games_api.py
+++ b/firstvoices/backend/tests/test_apis/test_games_api.py
@@ -12,6 +12,7 @@ from backend.tests.factories import (
     get_site_with_member,
 )
 from backend.tests.test_apis.base_api_test import BaseSiteContentApiTest
+from backend.tests.utils import equate_list_content_without_order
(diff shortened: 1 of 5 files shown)


firstvoices/backend/views/games_views.py
```diff
@@ -0,0 +1,118 @@
...
+                .filter(
+                    site=site,
+                    visibility=site.visibility,
+                    exclude_from_games=False,
+                    exclude_from_kids=False,
+                    chars_length__exact=SOLUTION_LENGTH,
+                )
+                .exclude(title__contains=" ")
+                .order_by("custom_order")
+                .values_list("title", flat=True)
```
caforbes, line 101: I tested locally and found two issues with the resulting list:

1. The list should have only unique words - duplicates should be removed.
2. The words need to be provided to the game using base characters only (e.g. `split_chars_base` joined). This is because the titles might be using variations like uppercase, which won't be correctly processed by the game. (Example: alphabet `a b c` has variants `A B C`, and the dictionary has a word `Abcab`. this word should be in the config as `abcab` not `Abcab`.)

firstvoices/backend/tests/test_apis/test_games_api.py
```diff
@@ -0,0 +1,138 @@
...
+        new_entry = DictionaryEntryFactory(
+            site=self.site, visibility=Visibility.PUBLIC, title="ccccc"
+        )
+
+        # Manually clearing the cache
+        caches[CACHE_KEY_WORDSY].clear()
+
+        # Fetching response again, the new entry should be now present in the words list
+        # since cache is cleared, so the db should be accessed again
+        response_new = self.client.get(self.get_list_endpoint(site_slug=self.site.slug))
```
caforbes, line 136: when testing locally, I had a problem getting a deleted word to delete from the list, even though I had cleared the cache. Maybe double check by adding tests for deletion, or changing visibility?

firstvoices/backend/tests/test_apis/test_games_api.py
```diff
@@ -0,0 +1,138 @@
...
+        assert response.status_code == 200
+        response_data = json.loads(response.content)
+
+        assert response_data["orthography"] == ["a", "b", "c"]
+        assert response_data["words"] == []
+        assert response_data["validGuesses"] == []
+        assert response_data["solution"] == ""
+
+    @pytest.mark.django_db
+    def test_all_invalid_words(self):
```
caforbes, line 48: just double checking, if there are not enough words to play the wordsy game (e.g. too few possibles words + solutions), is this handled on the FE? or does it need to be handled on the BE?

caforbes (commented):
A few questions and things to tweak, but overall it looks great and is really close!

NolanVH (approved):
Aside from the few issues mentioned by Clarissa this is looking great. No need for a re-review after those issues are addressed so I will approve now.

firstvoices/backend/tests/test_apis/test_games_api.py
sahil97: Yes, the enough words validation is handled on the FE side.

firstvoices/backend/views/games_views.py
```diff
@@ -0,0 +1,122 @@
...
+                    site=site,
+                    visibility=site.visibility,
+                    exclude_from_games=False,
+                    exclude_from_kids=False,
+                    chars_length__exact=SOLUTION_LENGTH,
+                )
+                .exclude(title__contains=" ")
+                .order_by("title", "custom_order")
+                .distinct("title")
+                .values_list("split_chars_base", flat=True)
```
caforbes, line 102: Distinct titles isn't enough to prevent duplicates here. For example if `Abcab` and `abcab` are both in the db then their shared base word `abcab` will show up twice here. You might need to look for distinct `split_chars_base` or do a final filter for duplicates.

caforbes (approved):
Found just one small fix!

firstvoices/backend/views/games_views.py
sahil97: Right, nice catch. I will add that change.

PR #763, opened 2024-01-17
Fw 5349 rebuild mtd index when saving a public dictionary entry

sahil97:
### Description of Changes
- Added signals to rebuild the mtd language index when relevant models are updated. The signals are placed at `/models/signals/mtd_signals.py`
- There was no need to add additional signals for notes, translations and acknowledgements since if they are to be changed, the dictionary entry is saved, thus the index is rebuilt anyway.
- Added through signals for images and videos.
- Added separate signals for category models with relevant checks to prevent unnecessary rebuilding.

### Checklist
- [x] Tests have been updated / created if applicable~

### DIFF
--- a/firstvoices/backend/apps.py
+++ b/firstvoices/backend/apps.py
@@ -4,3 +4,8 @@ from django.apps import AppConfig
 class BackendConfig(AppConfig):
     default_auto_field = "django.db.models.BigAutoField"
     name = "backend"
+
+    def ready(self):
+        import backend.models.signals  # noqa F401
+
+        # Search signals to be added later
--- a/firstvoices/backend/models/__init__.py
+++ b/firstvoices/backend/models/__init__.py
@@ -22,7 +22,6 @@ from .media import Image  # noqa F401
 from .mtd import MTDExportFormat  # noqa F401
 from .page import SitePage  # noqa F401
 from .part_of_speech import PartOfSpeech  # noqa F401
-from .signals.mtd_signals import *  # noqa F401 - MTD Index signals
 from .sites import Membership, Site  # noqa F401
 from .song import Lyric, Song  # noqa F401
 from .story import Story, StoryPage  # noqa F401
--- a/firstvoices/backend/models/__init__.py
+++ b/firstvoices/backend/models/__init__.py
@@ -22,6 +22,7 @@ from .media import Image  # noqa F401
 from .mtd import MTDExportFormat  # noqa F401
 from .page import SitePage  # noqa F401
 from .part_of_speech import PartOfSpeech  # noqa F401
+from .signals.mtd_signals import *  # noqa F401
 from .sites import Membership, Site  # noqa F401
 from .song import Lyric, Song  # noqa F401
 from .story import Story, StoryPage  # noqa F401
--- a/firstvoices/backend/models/__init__.py
+++ b/firstvoices/backend/models/__init__.py
@@ -22,7 +22,7 @@ from .media import Image  # noqa F401
 from .mtd import MTDExportFormat  # noqa F401
 from .page import SitePage  # noqa F401
 from .part_of_speech import PartOfSpeech  # noqa F401
-from .signals.mtd_signals import *  # noqa F401
+from .signals.mtd_signals import *  # noqa F401 - MTD Index signals
 from .sites import Membership, Site  # noqa F401
 from .song import Lyric, Song  # noqa F401
 from .story import Story, StoryPage  # noqa F401
--- a/firstvoices/backend/models/signals/mtd_signals.py
+++ b/firstvoices/backend/models/signals/mtd_signals.py
@@ -1,5 +1,4 @@
 from django.db import transaction
-from django.db.models import Count
 from django.db.models.signals import post_delete, post_save
 from django.dispatch import receiver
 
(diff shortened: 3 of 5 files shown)


firstvoices/backend/models/signals/mtd_signals.py
```diff
@@ -0,0 +1,47 @@
...
+            (site_slug,),
+            link_error=link_error_handler.s(),
+        )
+    )
+
+
+@receiver(post_save, sender=DictionaryEntry)
+@receiver(post_delete, sender=DictionaryEntry)
+def request_update_mtd_index(sender, instance, **kwargs):
+    if instance.visibility == Visibility.PUBLIC:
```
sarahfirstvoices, line 24: Does this handle cases where an entry is changed from public to members and vice versa?

firstvoices/backend/models/signals/mtd_signals.py
```diff
@@ -0,0 +1,47 @@
...
+@receiver(post_delete, sender=Category)
+def request_update_mtd_index_category_ops(sender, instance, **kwargs):
+    # Check if there are any relevant dictionary entries affected
+    # relevant dictionary entries are public and have at-least one translation
+    relevant_dictionary_entries = DictionaryEntry.objects.filter(
+        categories=instance, visibility=Visibility.PUBLIC
+    )
+    relevant_dictionary_entries = relevant_dictionary_entries.annotate(
+        translation_count=Count("translation_set")
+    ).filter(translation_count__gt=0)
```
sarahfirstvoices, line 38: I think you can also do this with something like relevant_dictionary_entries.exclude(translation_set=None)

firstvoices/backend/models/signals/mtd_signals.py
```diff
@@ -0,0 +1,47 @@
...
+    if len(relevant_dictionary_entries) > 0:
```
sarahfirstvoices, line 39: Probably safer to use .count() instead of len() here, since we don't need any of the dictionary entry values

firstvoices/backend/models/signals/mtd_signals.py
```diff
@@ -0,0 +1,47 @@
...
+        rebuild_mtd_index(instance.site.slug)
+
+
+@receiver(m2m_changed, sender=DictionaryEntry.related_audio.through)
+@receiver(m2m_changed, sender=DictionaryEntry.related_images.through)
+def request_update_mtd_index_audio_ops(sender, instance, **kwargs):
+    if instance.visibility == Visibility.PUBLIC:
+        rebuild_mtd_index(instance.site.slug)
```
sarahfirstvoices, line 47: Good catch on the categories and related media!

firstvoices/backend/models/signals/mtd_signals.py
```diff
@@ -0,0 +1,47 @@
...
+        translation_count=Count("translation_set")
+    ).filter(translation_count__gt=0)
+    if len(relevant_dictionary_entries) > 0:
+        rebuild_mtd_index(instance.site.slug)
+
+
+@receiver(m2m_changed, sender=DictionaryEntry.related_audio.through)
+@receiver(m2m_changed, sender=DictionaryEntry.related_images.through)
+def request_update_mtd_index_audio_ops(sender, instance, **kwargs):
+    if instance.visibility == Visibility.PUBLIC:
```
sarahfirstvoices, line 46: Is this checking the visibility of the through model instance or the dictionary entry?

firstvoices/backend/models/__init__.py
```diff
@@ -22,6 +22,7 @@
 from .mtd import MTDExportFormat  # noqa F401
 from .page import SitePage  # noqa F401
 from .part_of_speech import PartOfSpeech  # noqa F401
+from .signals.mtd_signals import *  # noqa F401 - MTD Index signals
```
sarahfirstvoices, line 25: Why do these need to be imported here?

sarahfirstvoices:
When we meet up, let's talk about options for testing these new signals

firstvoices/backend/models/signals/mtd_signals.py
sahil97: Yes this is checking the visibility of the dictionary entry.

firstvoices/backend/models/__init__.py
sahil97: As discussed, this import of signals has now been moved to the `apps.py` in the `ready()` method of the `BackendConfig`.

firstvoices/backend/models/signals/mtd_signals.py
sahil97: Yep, this worked. Thanks for the suggestion.

firstvoices/backend/models/signals/mtd_signals.py
sahil97: Good catch, it previously did not. I have now added pre_save and modified the post_save signal to check the previous visibility and rebuild index on the basis of that. That should cover the public to non-public visibility cases.

firstvoices/backend/apps.py
```diff
@@ -4,3 +4,8 @@
 class BackendConfig(AppConfig):
     default_auto_field = "django.db.models.BigAutoField"
     name = "backend"
+
+    def ready(self):
+        import backend.models.signals  # noqa F401
+
```
sarahfirstvoices, line 10: Thanks for fixing this

sarahfirstvoices (approved):
Tests look great. Thanks!

PR #765, opened 2024-01-18
Fw 5381 import embedded video links

caforbes:
### Description of Changes
Added a field for related video links to the Dictionary, Song, and Story imports. Added testing for video links and reorganized which migration resource tests are skipped.

Note: error-raising is removed from the import script so that we gracefully handle imports onto site data that may have since been deleted (it will skip without failing). And because we are slowly removing support for the import process.

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/management/commands/import_csv_data.py
+++ b/firstvoices/backend/management/commands/import_csv_data.py
@@ -160,7 +160,7 @@ def run_import():
 
             # raise errors to halt the import if an issue occurs
             try:
-                result = resource.import_data(dataset=table, raise_errors=True)
+                result = resource.import_data(dataset=table, raise_errors=False)
                 logger.info(
                     " ".join(
                         [f"{type}: {total}" for type, total in result.totals.items()]
--- a/firstvoices/backend/management/commands/import_csv_data.py
+++ b/firstvoices/backend/management/commands/import_csv_data.py
@@ -61,7 +61,6 @@ def run_import():
     Locally downloads import files from AWS, then matches them with the right
     backend model resource to import with. Deletes the database, then imports
     model-by-model in a specified order into the clean space.
-    Halts the process when encountering any data/validation error.
 
     Run with:
         python manage.py import_csv_data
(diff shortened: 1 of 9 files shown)


firstvoices/backend/management/commands/import_csv_data.py
```diff
@@ -160,7 +159,7 @@ def run_import():
 
             # raise errors to halt the import if an issue occurs
             try:
-                result = resource.import_data(dataset=table, raise_errors=True)
+                result = resource.import_data(dataset=table, raise_errors=False)
                 logger.info(
```
sarahfirstvoices, line 163: Is the idea that we will only export the embedded video data? Is there anything we need to change in the import script to avoid overwriting or duplicating existing data for other models? (I'm sure you have thought about this; I'm just out of the loop.)

caforbes: The way I set it up is:
* The exporter produces CSVs for words/phrases/songs/stories, but only has rows for content that had an embedded video on it.
* The CSV only contains the ID field, the Site ID field, and the related video links field, so only those fields will be changed.
* The importer will match content on its ID. Only the site-id and related-video-links will be updated (I think it is a safe assumption that the site IDs will be the same). And the `last_modified` date will be updated to now.
* If the ID can't be matched to any content, e.g. because it's since been deleted, the importer would first try to create it, but would fail because it doesn't have enough fields, like a title. With the change in this line the script will just pass over those rows instead of crashing.

caforbes: I also tested this on the local copy of the migration I had, and didn't run into any overwriting issues. Others are welcome to test, and we should check on preprod. (unfortunately none of the sites with embedded videos were populated on dev so it will be hard to test there.)

bronson-c (approved):
LGTM! Thank you for explaining to error raising removal. I was a little wary but it seems like it's not possible for data being lost and it will be tested on preprod 👍

PR #787, opened 2024-01-29
FW-4627 Media update APIs original field read-only fix

sahil97:
### Description of Changes
- updated media APIs to mark the original field as read-only for PUT/PATCH requests
- Added json parser support to all media views. Primarily the use case is for setting the `speakers` to an empty list for audio files.
- Refactored tests to also test for the above cases

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/serializers/media_serializers.py
+++ b/firstvoices/backend/serializers/media_serializers.py
@@ -104,6 +104,13 @@ class MediaSerializer(ExternalSiteContentUrlMixin, serializers.ModelSerializer):
         file.save()
         return file
 
+    def get_fields(self, *args, **kwargs):
+        fields = super().get_fields(*args, **kwargs)
+        request = self.context.get("request", None)
+        if request and getattr(request, "method", None) in ["PUT", "PATCH"]:
+            fields["original"].read_only = True
+        return fields
+
     class Meta:
         fields = base_id_fields + (
             "description",
--- a/firstvoices/backend/tests/test_apis/base_api_test.py
+++ b/firstvoices/backend/tests/test_apis/base_api_test.py
@@ -7,6 +7,7 @@ from rest_framework.reverse import reverse
 from rest_framework.test import APIClient
 
 from backend.models.constants import AppRole, Role, Visibility
+from backend.models.media import Audio, Image, Video
 from backend.tests import factories
 
 
@@ -696,7 +697,13 @@ class SiteContentUpdateApiTestMixin:
 
         expected_data = self.add_expected_defaults(data)
         self.assert_updated_instance(expected_data, self.get_updated_instance(instance))
-        self.assert_update_response(expected_data, response_data)
+
+        # We need to pass instance to verify that the original is not modified
+        # in case of media files
+        if self.model in [Audio, Image, Video]:
+            self.assert_update_response(instance, expected_data, response_data)
+        else:
+            self.assert_update_response(expected_data, response_data)
 
 
 class ControlledSiteContentUpdateApiTestMixin:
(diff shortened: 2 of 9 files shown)


NolanVH (approved):
LGTM!

sarahfirstvoices:
Not being able to test via Postman is a problem, and a good indication that we are doing something non-standard.

The usual way to set a null value with a multipart/form-data encoding is to not send that value at all. That could be fine for a PUT API, but will never work for PATCH.

Let me think about this for a bit. Since the file itself is read-only, we only need the form-data encoding for the POST. I'm going to see if it's practical to use a different encoding for the update APIs.

bronson-c (approved):
LGTM!

firstvoices/backend/tests/test_apis/base_api_test.py
```diff
@@ -673,7 +674,11 @@ def test_update_success_200(self):
         assert response_data["id"] == str(instance.id)
 
         self.assert_updated_instance(data, self.get_updated_instance(instance))
-        self.assert_update_response(data, response_data)
+
+        if self.model in [Audio, Image, Video]:
```
sarahfirstvoices, line 678: Why is this type check necessary? The base class should not need to know about the specific types that are using it

sarahfirstvoices: You should be able to override assert_update_response in the base media test, I think

firstvoices/backend/tests/test_apis/base_api_test.py
```diff
@@ -696,7 +701,13 @@ def test_update_with_nulls_success_200(self):
...
         expected_data = self.add_expected_defaults(data)
         self.assert_updated_instance(expected_data, self.get_updated_instance(instance))
-        self.assert_update_response(expected_data, response_data)
+
+        # We need to pass instance to verify that the original is not modified
+        # in case of media files
+        if self.model in [Audio, Image, Video]:
+            self.assert_update_response_media(instance, expected_data, response_data)
+        else:
+            self.assert_update_response(expected_data, response_data)
```
sarahfirstvoices, line 710: Same comment here about overriding it in a subclass rather that putting a type check in the parent

firstvoices/backend/serializers/media_serializers.py
```diff
@@ -104,6 +104,13 @@ def create_file(self, validated_data, filetype):
         file.save()
         return file
 
+    def get_fields(self, *args, **kwargs):
```
sahil97, line 107: To mark the "original" fields as read-only for all media types.

firstvoices/backend/tests/test_apis/base_api_test.py
```diff
@@ -654,11 +654,15 @@ def test_update_instance_missing_404(self):
...
 
     @pytest.mark.django_db
-    def test_update_success_200(self):
+    def perform_success_request(self, data_with_nulls=False):
         site = self.create_site_with_app_admin(Visibility.PUBLIC)
 
         instance = self.create_minimal_instance(site=site, visibility=Visibility.PUBLIC)
-        data = self.get_valid_data(site)
+
+        if data_with_nulls:
```
sarahfirstvoices, line 662: Rather than passing in a flag, why not just pass in the data?

firstvoices/backend/tests/test_apis/base_api_test.py
```diff
@@ -654,11 +654,15 @@ def test_update_instance_missing_404(self):
         assert response.status_code == 404
 
     @pytest.mark.django_db
```
sarahfirstvoices, line 656: This isn't a test case so I don't think it needs the mark

firstvoices/backend/tests/test_apis/base_api_test.py
```diff
@@ -654,11 +654,15 @@ def test_update_instance_missing_404(self):
...
-    def test_update_success_200(self):
+    def perform_success_request(self, data_with_nulls=False):
```
sarahfirstvoices, line 657: Could we name this perform_detail_request instead? Nothing guarantees it will succeed ;)

sahil97: that is correct. I will add this change.

PR #790, opened 2024-02-01
FW-5432 Hide hidden sites in site APIs

bronson-c:
### Description of Changes
- Hides hidden sites from the sites list API, regardless of visibility
- Allows access to detail view of hidden sites only if the user is staff or a team member of the hidden site

### Checklist
- [x] ~~Team Postman workspace has been updated if applicable~~

- [ ] ~~Pull the branch and test locally~~

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/permissions/predicates/base.py
+++ b/firstvoices/backend/permissions/predicates/base.py
@@ -1,4 +1,4 @@
-from rules import predicate, Predicate
+from rules import Predicate, predicate
 
 from backend.models.constants import AppRole, Role, Visibility
 from backend.permissions.utils import get_app_role, get_site_role
@@ -98,10 +98,16 @@ def is_superadmin(user, obj):
     return get_app_role(user) == AppRole.SUPERADMIN
 
 
+@predicate
+def is_hidden_obj(user, obj):
+    return obj.is_hidden
+
+
 #
 # site and app role combos
 is_at_least_language_admin = Predicate(
-    has_language_admin_membership | is_at_least_staff_admin, name="is_at_least_language_admin"
+    has_language_admin_membership | is_at_least_staff_admin,
+    name="is_at_least_language_admin",
 )
 
 #
--- a/firstvoices/backend/permissions/predicates/base.py
+++ b/firstvoices/backend/permissions/predicates/base.py
@@ -99,7 +99,7 @@ def is_superadmin(user, obj):
 
 
 @predicate
-def is_hidden_obj(user, obj):
+def is_hidden_site(user, obj):
     return obj.is_hidden
 
 
--- a/firstvoices/backend/permissions/predicates/view.py
+++ b/firstvoices/backend/permissions/predicates/view.py
@@ -26,6 +26,6 @@ has_visible_site = Predicate(
     name="has_visible_site",
 )
 
-can_view_hidden_site = Predicate(
-    base.is_at_least_staff_admin | base.has_team_access, name="can_view_hidden_site"
+has_hidden_site = Predicate(
+    base.is_at_least_staff_admin | base.has_team_access, name="has_hidden_site"
 )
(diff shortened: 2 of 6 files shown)


firstvoices/backend/views/sites_views.py
```diff
@@ -72,6 +74,13 @@ class SiteViewSet(FVPermissionViewSetMixin, ModelViewSet):
     pagination_class = None
     serializer_class = SiteDetailWriteSerializer
 
+    def initial(self, *args, **kwargs):
+        if self.action == "retrieve":
+            obj = self.get_object()
+            if obj.is_hidden and not can_view_hidden_site(self.request.user, obj):
```
sarahfirstvoices, line 80: Rather than defining this here, we should add this predicate to the permission rule that is defined on the site model. Otherwise we are going to get confused about what permission is being applied where

firstvoices/backend/permissions/predicates/view.py
```diff
@@ -25,3 +25,7 @@
     ),
     name="has_visible_site",
 )
+
+can_view_hidden_site = Predicate(
+    base.is_at_least_staff_admin | base.has_team_access, name="can_view_hidden_site"
+)
```
sarahfirstvoices, line 31: This is so readable, I love it.

sarahfirstvoices: I think once you use this on the model permissions, you will need a matching permission filter defined for it as well

firstvoices/backend/permissions/predicates/base.py
```diff
@@ -98,10 +98,16 @@ def is_superadmin(user, obj):
     return get_app_role(user) == AppRole.SUPERADMIN
 
 
+@predicate
+def is_hidden_obj(user, obj):
+    return obj.is_hidden
```
sarahfirstvoices, line 103: Will this cause errors if this predicate is used on a model that doesn't have an is_hidden field? If so, I think the name needs to be more specific, like is_hidden_site

firstvoices/backend/tests/test_apis/test_sites_api.py
```diff
@@ -941,3 +941,54 @@ def test_patch_success_200(self):
...
+        self.client.force_authenticate(user=user)
+        factories.SiteFactory.create(visibility=Visibility.PUBLIC, is_hidden=True)
+
+        response = self.client.get(self.get_list_endpoint())
+        response_data = json.loads(response.content)
+        assert len(response_data) == 0
+        assert response_data == []
+
+    @pytest.mark.django_db
+    def test_hidden_site_detail_403(self):
```
sarahfirstvoices, line 957: I think we need one more test like this one, but to prove that users with the MEMBER role also get a 403 on the detail api for a hidden site

firstvoices/backend/tests/test_apis/test_my_sites_api.py
```diff
@@ -125,3 +125,22 @@ def test_language_admin_role(self):
...
+    @pytest.mark.django_db
+    @pytest.mark.parametrize(
+        "role", [Role.MEMBER, Role.ASSISTANT, Role.EDITOR, Role.LANGUAGE_ADMIN]
+    )
+    def test_hidden_sites_visible(self, role):
+        user = factories.UserFactory.create()
+        hidden_site = factories.SiteFactory.create(
+            visibility=Visibility.PUBLIC, is_hidden=True
+        )
+        factories.MembershipFactory.create(user=user, site=hidden_site, role=role)
```
sarahfirstvoices, line 138: This API needs to use the permission you created-- users with the MEMBER role should not be able to see hidden sites via this API

firstvoices/backend/views/sites_views.py
```diff
@@ -213,3 +213,6 @@ def get_queryset(self):
             )
             .order_by(Upper("site__title"))
         )
+
+        queryset = queryset.exclude(Q(site__is_hidden=True) & Q(role__lte=Role.MEMBER))
```
bronson-c, line 217: This code prevents hidden sites from appearing to members in both detail and list view

sarahfirstvoices (approved):

PR #797, opened 2024-02-06
Fw-5298, signals for updating language index

sarahfirstvoices:
### Description of Changes
* add signals to asynchronously update language index when Language or Site instances are added, edited, or removed
* split DocumentManager out of IndexManager to better allow an index to cover multiple models
* switch the the standard way of importing signals

### Checklist

### Additional Notes
N/A


firstvoices/backend/search/indexing/base.py
```diff
@@ -0,0 +1,229 @@
...
+                e, type(cls.model).__name__, instance_id
+            )
+
+    @classmethod
+    def should_be_indexed(cls, instance):
+        """Subclasses can override for custom indexing conditions, such as visibility."""
+        return True
+
+    @classmethod
+    def sync_in_index(cls, instance_id):
```
sahil97, line 182: Great idea.

firstvoices/backend/tests/test_apis/test_languages_api.py
```diff
@@ -397,10 +397,13 @@ def test_search_mixed_sites_and_languages(self, db, mock_search_query_execute):
     def assert_language_response(self, language, language_response):
         assert language_response["language"] == language.title
         assert language_response["languageCode"] == language.language_code
+        assert language_response["id"] == str(language.id)
 
     def assert_more_sites_response(self, more_response):
-        assert more_response["language"] == "More FirstVoices Sites"
+        assert more_response["language"] == ""
```
sahil97, line 403: I guess this is related to the MoreSitesResponseSerializer which was renamed to LanguagePlaceholderSerializer. I believe we can rename this as well.

firstvoices/backend/search/indexing/base.py
```diff
@@ -0,0 +1,229 @@
...
+        es = connections.get_connection()
+        current_index = cls._get_current_index(es)
+        new_index = cls._create_new_write_index()
+
+        try:
+            cls._add_all(es)
+
+        except Exception as e:
+            # If we are not able to complete the new index, delete it and leave the current one as read + write alias
+            logger = logging.logger
```
sahil97, line 33: I believe this line can be removed

firstvoices/backend/search/indexing/base.py
```diff
@@ -0,0 +1,229 @@
...
+from firstvoices.settings import ELASTICSEARCH_DEFAULT_CONFIG
+
+
+class IndexManager:
+    index = ""
+    document_managers = []
+
+    @classmethod
+    def rebuild(cls):
+        logger = logging.logger
```
sahil97, line 21: I like the approach here, but I'm not sure why, but this instantiates the logger again. So we see statements twice in the logs. 

One way I tried that worked locally was, 
adding `logger.handlers.clear()` right after initializing the logger in `logging.py` or after instantiating the logger here.
Or we can import logging and rename the backend.search.logging something else/import as something else, and then grab the logger again `logger = logging.getLogger(ELASTICSEARCH_LOGGER)` and the log from there.

sarahfirstvoices: I poked around and I think the root issue is that the name of the elasticsearch logger was not following the package hierarchy with its naming, so it was propagating strangely to the root logger. I renamed it using the __name__ of the search logging module and the duplication went away. I can update our Sentry saved search to catch these logs

sahil97: Sounds good. Nice catch and thanks for figuring that out.

sarahfirstvoices: Thanks for noticing the dupes!

PR #844, opened 2024-03-20
FW-5417 more wordsy guesses

caforbes:
### Description of Changes

Refactoring Wordsy calculations to fix restrictions on the possible solution words, and create a separate, larger list of guessable words. Uses a less restrictive way of splitting characters  with regex - though this takes some time to process (on a site with ~13k words, took ~5 seconds locally).

Also includes minor refactors of the custom character flag.

### Checklist
- Insomnia workspace has been updated if applicable

### Additional Notes
To test locally, your data will need to be processed for games (have the character splitting data stored).

I'm not sure if this is too detailed/lengthy a process to be feasible. I could return to the original splitting method which could be done with only 1 query, sacrificing possible guesses for more faster speed. Thoughts?

### DIFF
--- a/firstvoices/backend/views/games_views.py
+++ b/firstvoices/backend/views/games_views.py
@@ -14,6 +14,8 @@ from rest_framework.response import Response
 from rest_framework.viewsets import GenericViewSet
 
 from backend.models import Character, DictionaryEntry
+from backend.models.dictionary import TypeOfDictionaryEntry
+from backend.utils.character_utils import UNKNOWN_FLAG
 from backend.views.api_doc_variables import site_slug_parameter
 from backend.views.base_views import SiteContentViewSetMixin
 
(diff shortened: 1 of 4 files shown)


firstvoices/backend/views/games_views.py
```diff
@@ -30,6 +33,67 @@ def get_wordsy_solution_seed(num_words):
...
+        .values_list("split_chars_base", flat=True)
+    )
+    solutions = []
+    for split_chars in words_qs.iterator():
+        title_as_base_chars = "".join(split_chars)
+        solutions.append(title_as_base_chars)
+    return solutions
+
+
+def get_wordsy_possible_guesses(site) -> list:
```
caforbes, line 66: This is the added section that searches for a different collection of possible guessable words, using different possible splitting methods. It is time consuming (~5sec for ~13000 entries) but the result is cached the first time the endpoint is hit.

If this is too slow, I could swap it out to use the original splitting method which would be much faster. And we could save this option (or a better one) for a future improvement in the async direction, maybe relying on a scheduled job instead.

sarahfirstvoices: Since the pool of valid guesses only changes if a word is added or edited, I wonder about keeping it up-to-date the same way we deal with the character splitting. If we stored something like is_valid_wordsy_solution and is_valid_wordsy_guess when we saved a word, we'd have to run recalculate dictionary once but then we could retrieve the valid words with a very fast lookup.

caforbes: I didn't think of this, but interesting idea. A possible drawback is that it wouldn't adjust well if we want the number of letters in the Wordsy game to be flexible based on difficulty or language (we would probably want to stick with 5 letters, period).

PR #850, opened 2024-03-28
FW-5433 Handle missing original file in media indexing

bronson-c:
### Description of Changes
Adds a conditional to the MediaDocumentManager to handle the case where the original file was missing and a matching test.

### Checklist

### Additional Notes
Currently the original file can be `None` but a separate ticket has been created to determine if that should still be possible. (See https://firstvoices.atlassian.net/browse/FW-5676)

### DIFF
--- a/firstvoices/backend/search/indexing/media_index.py
+++ b/firstvoices/backend/search/indexing/media_index.py
@@ -13,6 +13,13 @@ class MediaDocumentManager(DocumentManager):
     @classmethod
     def create_index_document(cls, instance):
         """Returns a MediaDocument populated for the given media instance."""
+
+        # Handle the case where the media instance has no original file
+        if instance.original:
+            instance_filename = instance.original.content.name
+        else:
+            instance_filename = None
+
         return cls.document(
             document_id=str(instance.id),
             document_type=type(instance).__name__,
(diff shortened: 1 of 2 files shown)


firstvoices/backend/search/indexing/media_index.py
```diff
@@ -13,6 +13,13 @@ class MediaDocumentManager(DocumentManager):
     @classmethod
     def create_index_document(cls, instance):
         """Returns a MediaDocument populated for the given media instance."""
+
+        # Handle the case where the media instance has no original file
+        if instance.original:
```
sahil97, line 18: Do we want to index entries without a file present ? As if we index entries without the media, during search, the results may appear broken to the users.

bronson-c: Currently the media models allow the `original` field to be null, I believe for migration reasons. A ticket has been made to research whether we should change the model (https://firstvoices.atlassian.net/browse/FW-5676) but this PR assumes that the `original` can still be allowed to be null

sarahfirstvoices: This is a good question... I'm pretty sure the media dashboard is using the search API, so if we ever want to be able to remove these blank models we need to keep indexing them for now

PR #896, opened 2024-04-29
FW 5683/5684 Queue and Run Bulk Visibility Jobs

bronson-c:
### Description of Changes
- Adds the ability for superusers to queue bulk visibility jobs via a POST to the bulk visibility API.
- Visibility must only change by 1 level.
- Adds the bulk visibility job to celery, which changes all entries on a site and the site itself by one visibility level.
- Additionally the bulk visibility job pauses and resumes search indexing as appropriate, so as to not queue an excessive amount of search tasks.
- Queues the bulk visibility job via a post-save signal. 
- Adds appropriate tests for bulk visibility API and task.

### Checklist

### Additional Notes
Spec document: https://firstvoices.atlassian.net/wiki/spaces/FIR/pages/541949966/Tech+design+Publishing+launching+sites#Proposal-for-site-visibility-API


sahil97 (dismissed):
PR looks good.
One suggestion I have would be to add a check before initiating a job to verify that multiple jobs on the same site are not started at the same time to avoid any conflicts.
Ideally, a job should wait for any previous BulkVisibilityJob running against the same site to finish and then start the next job.

firstvoices/backend/models/signals/visibility_signals.py
```diff
@@ -0,0 +1,27 @@
...
+from django.dispatch import receiver
+
+from backend.models.jobs import BulkVisibilityJob, JobStatus
+from backend.tasks.visibility_tasks import bulk_visibility_change_job
+from firstvoices.celery import link_error_handler
+
+
+@receiver(post_save, sender=BulkVisibilityJob)
+def request_bulk_visibility_change(sender, instance, **kwargs):
+    """
```
sarahfirstvoices, line 12: Is a signal necessary here? Since we only want to start this job one time, it seems like it might be easier to just queue the job in a regular function call after saving the initial job instance.

firstvoices/backend/models/signals/visibility_signals.py
```diff
@@ -0,0 +1,27 @@
...
+    Starts the bulk visibility change job when a BulkVisibilityJob is created.
+    """
+    if BulkVisibilityJob.objects.filter(
+        status=JobStatus.STARTED, site=instance.site
+    ).exists():
```
sarahfirstvoices, line 17: A detail about the timing of this check: we need to check for running jobs when we start running the bulk_visibility_change_job task, not when we queue it.

firstvoices/backend/tasks/visibility_tasks.py
```diff
@@ -0,0 +1,71 @@
...
+from backend.models.jobs import BulkVisibilityJob, JobStatus
+from backend.models.sites import SiteFeature
+from backend.models.widget import SiteWidget
+from backend.search.tasks.site_content_indexing_tasks import (
+    sync_all_site_content_in_indexes,
+)
+
+
+@shared_task
+def bulk_visibility_change_job(job_instance_id):
```
sarahfirstvoices, line 14: Functions should have verb names, so let's call this something like start_visibility_job or bulk_change_visibility

firstvoices/backend/tasks/visibility_tasks.py
```diff
@@ -0,0 +1,71 @@
...
+    job.status = JobStatus.STARTED
+    job.save()
+
+    # Pause search indexing for the site the job is for
+    if site.sitefeature_set.filter(key="indexing_paused").exists():
+        indexing_paused = site.sitefeature_set.get(key="indexing_paused")
+        indexing_paused.is_enabled = True
+        indexing_paused.save()
+    else:
+        SiteFeature.objects.create(site=site, key="indexing_paused", is_enabled=True)
```
sarahfirstvoices, line 29: This would be a good use for get_or_create. Like:

feature = SiteFeature.objects.get_or_create(site=site, key="indexing_paused")
feature.is_enabled = True
feature.save()

firstvoices/backend/tasks/visibility_tasks.py
```diff
@@ -0,0 +1,71 @@
...
+            entries.update(visibility=job.to_visibility)
+            songs.update(visibility=job.to_visibility)
+            stories.update(visibility=job.to_visibility)
+            story_pages.update(visibility=job.to_visibility)
+            pages.update(visibility=job.to_visibility)
+            widgets.update(visibility=job.to_visibility)
+
+        # Change Site visibility
+        site.visibility = job.to_visibility
+        site.save()
```
sarahfirstvoices, line 50: I believe the site visibility change should also be in the transaction

firstvoices/backend/tasks/visibility_tasks.py
```diff
@@ -0,0 +1,71 @@
...
+    except Exception as e:
+        # if there’s an error during the updates:
+        # add the message to the batch visibility job instance and set the status to failed
+        job.status = JobStatus.FAILED
+        job.message = str(e)
+        job.save()
+
+        indexing_paused = site.sitefeature_set.get(key="indexing_paused")
```
sarahfirstvoices, line 58: Do you need to re-fetch the site feature each time, or can you just set the values and save it?

sarahfirstvoices (changes requested):
Functionality is there! I added some comments about organization and timing

firstvoices/backend/tasks/visibility_tasks.py
bronson-c: I changed it to just 1 reference at the beginning of the task, nice catch!

firstvoices/backend/tasks/visibility_tasks.py
```diff
@@ -44,27 +51,25 @@ def bulk_visibility_change_job(job_instance_id):
...
+        indexing_paused_feature[0].save()
         return
 
     # Resume search indexing for site, + reindex entire site
     sync_all_site_content_in_indexes(site)
-    indexing_paused = site.sitefeature_set.get(key="indexing_paused")
-    indexing_paused.is_enabled = False
-    indexing_paused.save()
+    indexing_paused_feature[0].is_enabled = False
+    indexing_paused_feature[0].save()
```
sarahfirstvoices, line 72: This is minor, but should we turn indexing back on before we run the full re-index? The re-indexing takes some time, and any edits made during that time would not get synced in the index.

sarahfirstvoices (approved):

sahil97 (approved):
LGTM !

PR #907, opened 2024-05-08
Fw 5735 dry-run batch upload to generate the importJobReport

sahil97:
### Description of Changes
- Added signal to start a dry-run task after an importjob has been created
- Added an AudienceMixin to add audience related fields in the resources
- Moved `get_sample_file` utility to common utils file in tests
- Updated previous tests related to DictionaryEntryResource to reflect the AudienceMixin change.

### Checklist

### Additional Notes
N/A


firstvoices/backend/resources/utils/import_export_widgets.py
```diff
@@ -68,3 +71,13 @@ def clean(self, value, row=None, **kwargs):
                 **{self.field: DUMMY_USER_EMAIL}, defaults={"email": DUMMY_USER_EMAIL}
             )
             return dummy_user
+
+
+class AudienceBooleanFieldWidget(Widget):
```
sahil97, line 76: Can also be renamed to `ToggleBooleanFieldWidget` if its reused in more places than the Audience flags.

firstvoices/backend/models/signals/import_job_signals.py
```diff
@@ -0,0 +1,10 @@
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from backend.models.import_jobs import ImportJob
+from backend.tasks.import_job_tasks import execute_dry_run_import
+
+
+@receiver(post_save, sender=ImportJob)
+def request_dry_run_import(sender, instance, **kwargs):
+    execute_dry_run_import(instance)
```
sarahfirstvoices, line 10: I don't think we want to run this every time a job is saved, right? Only when it is first created. I think we need a condition here.

This also appears to be running synchronously, rather than being queued. Is that right?

firstvoices/backend/resources/utils/import_export_widgets.py
sarahfirstvoices: Maybe ReverseBooleanFieldWidget since it uses the opposite value

firstvoices/backend/tests/test_resources/test_dictionary_resources.py
```diff
@@ -37,7 +37,7 @@ def build_table(data: list[str]):
         headers = [
             # these headers should match what is produced by fv-nuxeo-export tool
```
sarahfirstvoices, line 38: Oops, this comment is out of date

firstvoices/backend/resources/dictionary.py
```diff
@@ -21,14 +23,17 @@
...
     BaseResource,
     RelatedMediaResourceMixin,
     SiteContentResource,
 )
 from backend.resources.utils.import_export_widgets import ChoicesWidget
 
 
-class DictionaryEntryResource(SiteContentResource, RelatedMediaResourceMixin):
+class DictionaryEntryResource(
+    AudienceMixin, SiteContentResource, RelatedMediaResourceMixin
```
sarahfirstvoices, line 35: You can use ControlledSiteContentResource instead to get the standard visibility field

sarahfirstvoices (changes requested):
Only blocker for me is making sure the task is running async and only once.

firstvoices/backend/resources/utils/import_export_widgets.py
caforbes: I'd consider naming this `InvertedBooleanWidget` or similar to indicate the value flip. Since you're here it might also be valuable to add a `render` method to show how the db value would be exported to CSV (it would probably also need to be flipped).

Also, where user input is concerned we will need to accept a wider range of values, but that could be a separate ticket.

firstvoices/backend/tasks/import_job_tasks.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # Create an ImportJobReport for the run
+    report = ImportJobReport(
+        site=import_job_instance.site,
+        importjob=import_job_instance,
+        new_rows=result.totals["new"],
+        skipped_rows=result.totals["skip"],
+        error_rows=result.totals["error"],
+    )
+    report.save()
```
caforbes, line 43: Question: It seems possible for the user to enter data that would result in an update or deletion (e.g. if they include an ID field that matches an existing entry). Do we want to track the other possible row values in the report, or anywhere in this job? Or will we handle this possibility later?

firstvoices/backend/tests/factories/resources/import_job/test_invalid_dictionary_entries.csv
```diff
@@ -0,0 +1,6 @@
+﻿title,type,visibility,acknowledgement,include_in_games,include_on_kids_site
+Word 1,xyz,Public,123,TRUE,FALSE
+Phrase 1,phrase,Team,123,FALSE,TRUE
+,notAType,Public,123,FALSE,TRUE
```
caforbes, line 4: Row 3 doesn't have any obvious issue to me, while row 4 has two issues (empty title and notAType). Is that correct? just checking.

firstvoices/backend/tests/test_resources/test_dictionary_resources.py
caforbes, line 38: Since we're keeping the resources (with modifications) and tests, some of these comments should be removed.

If I can help with building automated test data, lmk.

caforbes (commented):
Just a few questions. Some might be issues for later tickets but wanted to raise them now.

firstvoices/backend/tests/factories/resources/import_job/test_invalid_dictionary_entries.csv
sahil97: Yeah, nice catch. That row is fine, I just left it there as a control.

firstvoices/backend/tasks/import_job_tasks.py
sahil97: Yes we have plans to identify the invalid column name and value later and then put them in the `identifier_field` and `identifier_value` fields of this model. As of now, the errors are just returned as is, for each row, and we are not parsing them for MVP.
Please let me know if we want to do that for MVP and I can create tickets for that.

firstvoices/backend/resources/utils/import_export_widgets.py
sahil97: Thank you for the suggestion. I have updated the name to `InvertedBooleanFieldWidget`. 
Also, I've added a ticket for the render method at https://firstvoices.atlassian.net/browse/FW-5842.

firstvoices/backend/tests/test_resources/test_dictionary_resources.py
sahil97: Yes, I have updated this comment, and I will be reaching out to you for help in building the test data once the dry-run base infra is complete on my end so we can test it better and catch all the edge cases. Thank you :)

firstvoices/backend/tasks/import_job_tasks.py
sarahfirstvoices: @caforbes Are you seeing a way in the current API for someone to cause an edit or delete instead of only adding rows?

sarahfirstvoices: I get it now :) We logged a ticket to enforce the import mode we are intended to use (create vs edit).

sarahfirstvoices (approved):

PR #950, opened 2024-06-20
FW-5940 CSV validation flow refactoring for import-job dry run

sahil97:
### Description of Changes
- Removed unused resources and their tests from DictionaryResource file.
- Added `accepted_columns, ignored_columns` to import job model and serializer.
- Added invalid rows to result total for an import job.
- Added validation for input categories during batch import.
- Added validation for duplicate headers in csv.
- Refactored csv headers validation during batch import to make it more scalable.

### Checklist

### Additional Notes
N/A


firstvoices/backend/resources/utils/import_export_widgets.py
```diff
@@ -105,13 +105,26 @@ def __init__(self, *args, **kwargs):
...
+                input_categories[column] = input_value
+
+        # If no categories provided, return
         if len(input_categories) == 0:
             return Category.objects.none()
 
-        return Category.objects.filter(title__in=input_categories)
+        # Validate categories
+        for column, input_value in input_categories.items():
+            category_lookup = Category.objects.filter(
```
bronson-c, line 122: [minor] Does this category lookup mean that the category in the csv file needs to match the title of an existing category 1:1? If so would making it case-insensitive have any benefit for the batch-importer?

bronson-c (approved):
Just one comment about a minor change that could have some benefit. Otherwise LGTM! 👍

firstvoices/backend/resources/utils/import_export_widgets.py
sarahfirstvoices: This is a good thought, but currently category uniqueness is case-sensitive. I managed to set up both "Animals" and "animals" as categories on one of the dev sites. 

Let's keep this import code case-sensitive, and I will log a ticket to consider making category uniqueness case insensitive.

firstvoices/backend/resources/utils/import_export_widgets.py
```diff
@@ -105,13 +105,26 @@ def __init__(self, *args, **kwargs):
 
     def clean(self, value, row=None, **kwargs):
         category_match_pattern = r"category[_2-5]*"
```
sarahfirstvoices, line 107: Just so we don't get confused in future about whether this is matching category names vs the names of category columns, could we please change the name to something like category_column_name_pattern?

firstvoices/backend/resources/utils/import_export_widgets.py
sahil97: Yes, in the tech design, it was mentioned to be case-sensitive as to conform with the db model so I kept it that way.
+1 for the ticket to make category case-insensitive.

firstvoices/backend/resources/utils/import_export_widgets.py
```diff
@@ -105,13 +105,26 @@ def __init__(self, *args, **kwargs):
...
             return Category.objects.none()
 
-        return Category.objects.filter(title__in=input_categories)
+        # Validate categories
+        for column, input_value in input_categories.items():
+            category_lookup = Category.objects.filter(
+                site__id=row["site"], title=input_value
+            )
+            if len(category_lookup) == 0:
+                raise ValidationError(f"Invalid category supplied in column {column}.")
```
sarahfirstvoices, line 126: Just to make sure I understand how these errors work-- this will lead to this row being skipped in the import but will not fail the whole import, right?

firstvoices/backend/tasks/import_job_tasks.py
```diff
@@ -11,6 +11,76 @@
...
+        prefix = "_".join(splits[:-1])
+        variation = splits[-1]
+    else:
+        prefix = input_header
+        variation = None
+
+    # Check if the prefix is a valid header
+    if prefix in VALID_HEADERS and variation and variation.isdigit():
+        variation = int(variation)
+        if variation < 1 or variation > 5:
```
sarahfirstvoices, line 50: It's not allowed to be 1 either, right? This should be <=1

sarahfirstvoices (approved):

firstvoices/backend/resources/utils/import_export_widgets.py
sahil97: Yes, that's right. And this error message will go to this row's errors list.

firstvoices/backend/tasks/import_job_tasks.py
sahil97: Oh right, nice catch!

bronson-c (approved):

PR #963, opened 2024-07-04
FW-5936 Run MTD sync tasks on a schedule

bronson-c:
### Description of Changes
- Removes MTD signals from dictionary entries and categories
- Adds a periodic task for checking if a site has had changes in the last 6 hours that should trigger a MTD reindex
- The periodic task calls an async MTD reindex if so
- The periodic task is created as a model on db startup

### Checklist

### Additional Notes
- Currently this task does not use the `BaseJob` model, but if the changes from our incident response determine that all celery tasks should use `BaseJob` I can add that functionality to this PR.


firstvoices/backend/apps.py
```diff
@@ -4,7 +4,3 @@
 class BackendConfig(AppConfig):
     default_auto_field = "django.db.models.BigAutoField"
     name = "backend"
-
-    def ready(self):
-        import backend.models.signals  # noqa F401
-        import backend.search.signals  # noqa F401
```
sarahfirstvoices, line 10: Is this intentional? This seems like it would disable all signals

firstvoices/backend/tasks/build_mtd_export_format_tasks.py
```diff
@@ -134,3 +138,50 @@ def build_index_and_calculate_scores(site_or_site_slug: str | Site, *args, **kwa
     MTDExportFormat.objects.filter(site=site).exclude(id=new_result.id).delete()
 
     return preview
+
+
+@shared_task(bind=True)
+def check_sites_for_mtd_sync(self):
+    try:
+        LOGGER.info("Starting MTD Sync check.")
```
sarahfirstvoices, line 146: todo: update this with the standard logging sahil is setting up

firstvoices/backend/tests/test_tasks/test_build_mtd.py
```diff
@@ -158,3 +164,115 @@ def test_old_results_removed(self, site):
...
+
+        result = check_sites_for_mtd_sync.apply()
+        assert result.state == "SUCCESS"
+        assert self.mocked_func.call_count == 1
+        self.mocked_func.assert_called_once_with(
+            (sites["site_one"].slug,), link_error=link_error_handler.s()
+        )
+
+    @pytest.mark.django_db
+    def test_single_site_updated_related_media(self, sites):
```
sarahfirstvoices, line 225: Please add a test case to verify we will pick it up when related media gets removed from a dictionary entry

firstvoices/backend/tasks/build_mtd_export_format_tasks.py
```diff
@@ -134,3 +138,50 @@ def build_index_and_calculate_scores(site_or_site_slug: str | Site, *args, **kwa
...
+        self.state = "STARTED"
+        sites = Site.objects.all()
```
sarahfirstvoices, line 148: We can filter this to only public sites

sarahfirstvoices: Which will be good, since we can no longer filter the dictionary entries based on visibility. (We could catch all the currently public ones that changed, but we would miss any formerly public ones that got made private.)

firstvoices/backend/tests/test_tasks/test_build_mtd.py
```diff
@@ -158,3 +164,115 @@ def test_old_results_removed(self, site):
...
+    def sites(self):
+        return {
+            "site_one": factories.SiteFactory.create(slug="site_one"),
+            "site_two": factories.SiteFactory.create(slug="site_two"),
+            "site_three": factories.SiteFactory.create(slug="site_three"),
+        }
+
+    @pytest.mark.django_db
+    def test_no_sites_eligible_to_sync(self):
+        result = check_sites_for_mtd_sync.apply()
```
sarahfirstvoices, line 186: Please add another case where only private sites have changes, and we don't queue the task

firstvoices/backend/apps.py
bronson-c: Ah the backend.models.signals was intentional but the search signals should still be there, I will re-add the line, thanks!

sarahfirstvoices (approved):

sahil97 (approved):

PR #969, opened 2024-07-10
FW-5935 Remove site feature signal syncing media indexes

bronson-c:
### Description of Changes
- Removes the signal tied to site features that re-indexed a site's media on create and delete
- Added synchronous media indexing that runs only when the site feature API is used
- Added tests to ensure the media sync does not occur when site features are created/modified/deleted directly in the django db
- Added tests to ensure the reindexing does occur when using the site feature API

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/views/site_feature_views.py
+++ b/firstvoices/backend/views/site_feature_views.py
@@ -2,6 +2,9 @@ from drf_spectacular.utils import OpenApiResponse, extend_schema, extend_schema_
 from rest_framework.viewsets import ModelViewSet
 
 from backend.models.sites import SiteFeature
+from backend.search.tasks.site_content_indexing_tasks import (
+    sync_all_media_site_content_in_indexes,
+)
 from backend.serializers.site_feature_serializers import SiteFeatureDetailSerializer
 from backend.views import doc_strings
 from backend.views.api_doc_variables import key_parameter, site_slug_parameter
@@ -37,7 +40,8 @@ from backend.views.base_views import FVPermissionViewSetMixin, SiteContentViewSe
         ],
     ),
     create=extend_schema(
-        description="Create a new feature flag for the site. Site feature keys cannot be changed after creation.",
+        description="Create a new feature flag for the site. Site feature keys cannot be changed after creation. "
+        "Triggers a reindex of all media content in the site.",
         responses={
             201: OpenApiResponse(
                 description=doc_strings.success_201,
@@ -50,7 +54,8 @@ from backend.views.base_views import FVPermissionViewSetMixin, SiteContentViewSe
         parameters=[site_slug_parameter],
     ),
     update=extend_schema(
-        description="Edit a feature flag for the specified site. Site feature keys cannot be changed after creation.",
+        description="Edit a feature flag for the specified site. Site feature keys cannot be changed after creation."
+        "Triggers a reindex of all media content in the site.",
         responses={
             200: OpenApiResponse(
                 description=doc_strings.success_200_edit,
@@ -66,7 +71,8 @@ from backend.views.base_views import FVPermissionViewSetMixin, SiteContentViewSe
         ],
     ),
     partial_update=extend_schema(
-        description="Edit a feature flag for the specified site. Any omitted fields will be unchanged.",
+        description="Edit a feature flag for the specified site. Any omitted fields will be unchanged."
+        "Triggers a reindex of all media content in the site.",
         responses={
             200: OpenApiResponse(
                 description=doc_strings.success_200_edit,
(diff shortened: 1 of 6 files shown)


firstvoices/backend/tasks/visibility_tasks.py
```diff
@@ -69,7 +69,7 @@ def bulk_change_visibility(job_instance_id):
     # Resume search indexing for site, + reindex entire site
     indexing_paused_feature[0].is_enabled = False
     indexing_paused_feature[0].save()
-    sync_all_site_content_in_indexes(site)
+    sync_all_site_content_in_indexes(site.id)
```
bronson-c, line 72: Note: this call to the `sync_all_media_site_content_in_indexes` task is used without apply_async as it is already in the `bulk_change_visibility` async task

sahil97: Not sure, but do we want to use a apply_async call for consistency in calling/logging async tasks ?

Logs from this call will be under the parent task's id, though we will still have the method name in the log for debugging.
Sample log format for reference:
`[timestamp: INFO/PoolWorker-X] backend.tasks.send_email_task[task-id]: Task started.`

sahil97 (dismissed):
Other from the minor doubt about calling the sync_all_site_content_in_indexes without apply_async, the rest of the code looks good. Thank you for making the changes.

firstvoices/backend/views/site_feature_views.py
```diff
@@ -37,7 +40,8 @@
         ],
     ),
     create=extend_schema(
-        description="Create a new feature flag for the site. Site feature keys cannot be changed after creation.",
+        description="Create a new feature flag for the site. Site feature keys cannot be changed after creation. "
+        "Triggers a reindex of all media content in the site.",
```
sarahfirstvoices, line 46: All of our APIs keep the content up to date in the search indices, so I don't think we need to mention this specifically in the docs.

firstvoices/backend/views/site_feature_views.py
```diff
@@ -109,3 +118,31 @@ def get_queryset(self):
...
+        # Once a site feature is created via the API, sync all media content in indexes
+        instance = serializer.save()
+        site = instance.site
+
+        # Queue the site media sync after updating the model
+        transaction.on_commit(
+            lambda: sync_all_media_site_content_in_indexes.apply_async(
+                (site.id,), link_error=link_error_handler.s()
+            )
+        )
```
sarahfirstvoices, line 132: I know we're only doing this twice, but I think it is low-level enough that we extract it into a function called something like request_sync_media_index_for_site . You could put that function in search.tasks.site_content_indexing_tasks -- we have similar ones in the index_manager_tasks module.

firstvoices/backend/tasks/visibility_tasks.py
bronson-c: adding apply_async for log consistency, thanks!

firstvoices/backend/search/tasks/site_content_indexing_tasks.py
```diff
@@ -11,25 +12,26 @@
...
+    logger = get_task_logger(__name__)
+    logger.info(ASYNC_TASK_START_TEMPLATE, f"site: {site_title}")
 
-    remove_all(DictionaryEntryDocumentManager, site.dictionaryentry_set.all())
-    remove_all(SongDocumentManager, site.song_set.all())
-    remove_all(StoryDocumentManager, site.story_set.all())
-    remove_all(AudioDocumentManager, site.audio_set.all())
-    remove_all(ImageDocumentManager, site.image_set.all())
-    remove_all(VideoDocumentManager, site.video_set.all())
+    remove_all(DictionaryEntryDocumentManager, site_content_ids["dictionaryentry_set"])
```
sarahfirstvoices, line 29: This needs to take the site id as the parameter and look up the rest at runtime, like it was before. This is in case the content has changed in between the task being queued and starting to run

sarahfirstvoices: Oh wait, I bet all the content has been deleted by then, right? Was this a fix for that?

firstvoices/backend/tasks/visibility_tasks.py
```diff
@@ -82,7 +82,7 @@ def bulk_change_visibility(job_instance_id):
     # Resume search indexing for site, + reindex entire site
     indexing_paused_feature[0].is_enabled = False
     indexing_paused_feature[0].save()
-    sync_all_site_content_in_indexes(site.id)
+    request_sync_all_site_content_in_indexes(site)
```
sarahfirstvoices, line 85: Is there any benefit to queuing this as a separate task? This is already running in the background

sarahfirstvoices: I do think we could probably save the job before running the reindex

sarahfirstvoices: We chatted in person. +1 for queuing this separately so we can track the task separately in the logs.

sarahfirstvoices (approved):

PR #984, opened 2024-07-24
FW-5688 Refactor Dictionary Cleanup to Use BaseJob Model

bronson-c:
### Description of Changes
- Fully refactors the dictionary cleanup API to be in line with bulk visibility functionality
- CustomOrderRecalculationsResults have been refactored to DictionaryCleanupJobs for consistency, they now inherit from the BaseJob class and have less unnecessary fields
- Merged the separate preview and custom order recalculate tasks into a single task, that checks if the job is a preview dynamically
- Added cancel and failure states and enhanced logging to the dictionary cleanup job
- Refactored the dictionary cleanup API to behave like bulk visibility, sending a POST created a model to track the task and store the results, past runs are now kept and the PENDING issue is resolved.
- Refactored the tests to match the new functionality.

### Checklist

### Additional Notes
N/A


firstvoices/backend/models/__init__.py
```diff
@@ -15,7 +15,7 @@
     ImportJobReport,
     ImportJobReportRow,
 )
-from .jobs import BulkVisibilityJob, CustomOrderRecalculationResult  # noqa F401
+from .jobs import BulkVisibilityJob, DictionaryCleanupJob  # noqa F401
 from .join_request import JoinRequest  # noqa F401
```
sarahfirstvoices, line 19: I'm very into this rename

firstvoices/backend/models/jobs.py
```diff
@@ -50,13 +50,11 @@ class Meta:
...
 
-    latest_recalculation_date = models.DateTimeField(auto_now_add=True)
-    latest_recalculation_result = models.JSONField()
-    task_id = models.CharField(max_length=255, null=True, blank=True)
+    cleanup_result = models.JSONField(blank=True, null=True)
     is_preview = models.BooleanField(default=True)
 
     def __str__(self):
-        return self.site.title + " - " + str(self.latest_recalculation_date)
+        return self.site.title + " - DictionaryCleanup - " + self.status
```
sarahfirstvoices, line 57: I think this __str__ needs something more individual than status, probably just the id. Like `self.site.title DictionaryCleanupJob self.id`

firstvoices/backend/serializers/job_serializers.py
```diff
@@ -84,38 +74,25 @@ def get_latest_recalculation_result(obj):
...
+class DictionaryCleanupPreviewJobSerializer(DictionaryCleanupJobSerializer):
+    url = SiteHyperlinkedIdentityField(
+        read_only=True, view_name="api:dictionary-cleanup-preview-detail"
     )
-    latest_recalculation_preview_date = serializers.DateTimeField(
-        source="latest_recalculation_date"
+    cleanup_preview_result = serializers.JSONField(
+        read_only=True,
+        source="cleanup_result",
     )
```
sarahfirstvoices, line 88: Since this is a dedicated serializer for preview jobs, could we set is_preview to True in here?

firstvoices/backend/serializers/job_serializers.py
```diff
@@ -36,28 +31,23 @@ class Meta:
...
-    site = serializers.StringRelatedField()
-    current_task_status = serializers.SerializerMethodField()
-    latest_recalculation_result = serializers.SerializerMethodField()
-
-    @staticmethod
-    @extend_schema_field(OpenApiTypes.STR)
-    def get_current_task_status(obj):
-        async_result = AsyncResult(obj.task_id)
-        return async_result.status
+class DictionaryCleanupJobSerializer(
```
sarahfirstvoices, line 34: Suggestion: could we use the same serializer for the real cleanup and the preview if we called the result a "result"?

firstvoices/backend/serializers/job_serializers.py
```diff
@@ -36,28 +31,23 @@ class Meta:
...
-            and obj.latest_recalculation_result
-        ):
-            unknown_character_count = obj.latest_recalculation_result.get(
+    def get_cleanup_result(obj):
+        if hasattr(obj, "cleanup_result") and obj.cleanup_result:
+            unknown_character_count = obj.cleanup_result.get(
                 "unknown_character_count", 0
             )
-            updated_entries = obj.latest_recalculation_result["updated_entries"]
+            updated_entries = obj.cleanup_result["updated_entries"]
```
sarahfirstvoices, line 50: Would it make sense to put separate fields for updated_entries and unknown_character_count etc directly on the model? Might make the API responses easier to read

sarahfirstvoices: I realize this is a mischief suggestion-- if other folks think it is a good idea we should log a separate ticket

firstvoices/backend/tasks/dictionary_cleanup_tasks.py
```diff
@@ -0,0 +1,112 @@
...
+from backend.models.jobs import JobStatus
+from backend.tasks.utils import ASYNC_TASK_END_TEMPLATE, ASYNC_TASK_START_TEMPLATE
+
+
+@shared_task
+def cleanup_dictionary(job_instance_id: str):
+    """
+    Calculates and returns the results of a custom order recalculation,
+    including the changes in custom order and title and the count of unknown characters.
+    Jobs marked as preview will not save the changes to the database.
```
sarahfirstvoices, line 14: To be extra clear, could we change the last sentence to "Jobs marked as preview will not update DictionaryEntries in the database"? I could imagine someone thinking job instances don't get saved if they are marked as preview

firstvoices/backend/views/dictionary_cleanup_views.py
```diff
@@ -0,0 +1,181 @@
...
+
+        # Queue the cleanup task after model creation
+        transaction.on_commit(
+            lambda: cleanup_dictionary.apply_async(
+                (instance.id,), link_error=link_error_handler.s()
+            )
+        )
+
+    @action(methods=["delete"], detail=False)
+    def clear(self, request, *args, **kwargs):
```
sarahfirstvoices, line 171: While we are making changes here, since this API changes multiple resources, in REST terms it should be a POST not a DELETE

firstvoices/backend/views/dictionary_cleanup_views.py
```diff
@@ -0,0 +1,181 @@
...
+            404: OpenApiResponse(description=doc_strings.error_404_missing_site),
+        },
+        parameters=[site_slug_parameter],
+    ),
+)
+class DictionaryCleanupJobViewSet(
+    SiteContentViewSetMixin,
+    FVPermissionViewSetMixin,
+    ModelViewSet,
+):
```
sarahfirstvoices, line 65: These 2 views are now so similar that I think we could refactor this further. If we pulled out the is_preview value to a class member, and if you take my suggestion of using the same serializer for both, I think DictionaryCleanupPreviewJobViewset could simply subclass DictionaryCleanupJobViewSet with a different is_preview value and api docs.

sarahfirstvoices (changes requested):
Looking really good so far! Most of my comments are about pushing it a little further :)

firstvoices/backend/serializers/job_serializers.py
bronson-c: Rather than logging a separate ticket, will add to dictionary cleanup memory changes.

firstvoices/backend/serializers/job_serializers.py
bronson-c: As per our discussion separate serializers are still required but to the difference in URL for the separate views. However the `DictionaryCleanupJobSerializer` has been updated such that the `DictionaryCleanupPreviewJobSerializer` only needs one line of code.

firstvoices/backend/views/dictionary_cleanup_views.py
bronson-c: As per our discussion separate views are still preferable, however, the `DictionaryCleanupJobViewSet` has been updated such that the `DictionaryCleanupPreviewViewSet` only needs documentation and two lines of code.

firstvoices/backend/serializers/job_serializers.py
bronson-c: Details added to this JIRA: https://firstvoices.atlassian.net/browse/FW-5933

firstvoices/backend/serializers/job_serializers.py
```diff
@@ -34,12 +34,19 @@ class Meta:
...
     is_preview = serializers.BooleanField(read_only=True)
     cleanup_result = serializers.SerializerMethodField(read_only=True)
 
+    def get_fields(self):
+        fields = super().get_fields()
+        # Dynamically assign the view_name to the url field
+        fields["url"] = SiteHyperlinkedIdentityField(
+            read_only=True, view_name=self.detail_view_name
+        )
+        return fields
```
sarahfirstvoices, line 48: Nice!

sarahfirstvoices (approved):
Looks great!

PR #1011, opened 2024-08-13
FW-5870 Handle bad image data in site data serializer

bronson-c:
### Description of Changes
- Slightly edits the `get_img` function of the DictionaryEntryDataSerializer to ignore images that do not have small image thumbnails. This allows bad data in the db to be ignored.

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/serializers/site_data_serializers.py
+++ b/firstvoices/backend/serializers/site_data_serializers.py
@@ -108,7 +108,8 @@ class DictionaryEntryDataSerializer(serializers.ModelSerializer):
         # NOTE: MTD currently only allows one image. As a heuristic, I'm selecting the first one.
         image = dictionaryentry.related_images.first()
 
-        if image is None:
+        # If no small image is available (possibly due to the wrong file type), return None
+        if image is None or image.small is None:
             return None
 
         return ImageDataSerializer(image).data["filename"]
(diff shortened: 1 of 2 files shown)


firstvoices/backend/serializers/site_data_serializers.py
```diff
@@ -108,7 +108,8 @@ def get_img(dictionaryentry):
         # NOTE: MTD currently only allows one image. As a heuristic, I'm selecting the first one.
         image = dictionaryentry.related_images.first()
 
-        if image is None:
+        # If no small image is available (possibly due to the wrong file type), return None
+        if image is None or image.small is None:
```
sahil97, line 112: minor: should we make it so, that if image.small is None, but the image is present (probably cause of some issue with thumbnail generation), we still send that as a backup ?

sahil97 (approved):
Other than a small question, LGTM !

firstvoices/backend/serializers/site_data_serializers.py
bronson-c: As per our discussion, if an image does not have a small thumbnail it likely is bad data in the first place, such as an unsupported file type. To avoid setting bad data as images in the app we will only return the small thumbnails.

PR #1018, opened 2024-08-21
FW-6069 + FW-6060 Add dynamic time check to check_sites_for_mtd_sync

bronson-c:
### Description of Changes
- Adds dynamic checking of the PeriodicTask object to automatically adjust the lastModified period checked by `check_sites_for_mtd_sync`
- Supports the interval and crontab PeriodicTask schedule types, any others will cause the period to default to 24 hours
- Adds the `croniter` library to calculate the time of the last run `check_sites_for_mtd_sync` task
- Adds tests for each new code path

### Checklist

### Additional Notes
N/A


firstvoices/backend/tasks/mtd_export_tasks.py
```diff
@@ -188,23 +190,27 @@ def check_sites_for_mtd_sync(self):
     try:
         self.state = "STARTED"
         sites = Site.objects.filter(visibility=constants.Visibility.PUBLIC)
-        six_hours_ago = timezone.now() - timedelta(hours=6)
+        delta = get_delta_for_mtd_check() or timedelta(hours=24)
+        last_modified_check_period = timezone.now() - delta
```
sarahfirstvoices, line 194: I can think of a few cases where this will have trouble, for example:

* We manually generated an MTD export a long time ago, then edited some data, and just now started running this task. The old changes will not be noticed because they didn't happen during the calculated time delta.

Could we adjust this logic to be more like: check the timestamp on the most recent MTD export, and if it is longer ago than some configurable minimum time (e.g., 1 hour), check for data that has changed since then? It is more meaningful to check the timestamp on the exported data rather than the last time the task ran, and probably easier as well.

sarahfirstvoices (changes requested):
If you want to chat about the timing logic, give me a call :)

bronson-c:
This PR now includes all the changes from https://github.com/First-Peoples-Cultural-Council/fv-be/pull/1027 as the latter PR includes some logic/functionality for this one. If changes are made to the latter PR they will be merged into this one as well.

firstvoices/backend/tasks/mtd_export_tasks.py
```diff
@@ -186,38 +206,57 @@ def check_sites_for_mtd_sync(self):
...
-                    | Q(related_videos__last_modified__gte=six_hours_ago)
-                )
-                .distinct()
-                .count()
+            completed_jobs = MTDExportJob.objects.filter(
+                site=site, status=JobStatus.COMPLETE
             )
+            if completed_jobs.exists():
+                last_export = completed_jobs.latest()
+                last_modified_check_period = last_export.created
```
sarahfirstvoices, line 217: Super minor, but could we please call this variable last_export_created ? I got confused because it is a timestamp not a period

sarahfirstvoices (approved):
I made one little suggestion about naming, but otherwise this looks great! Thanks for making the change about time detection

PR #1027, opened 2024-08-26
FW-6037 + FW-6043 Use baseJob logic in build_index_and_calculate_scores

bronson-c:
### Description of Changes
- Changes `MTDExportFormat` to use the `BaseJob` model
- Edits `build_index_and_calculate_scores` to cancel the job if another `build_index_and_calculate_scores` is running on the same site
- Adds a failure condition to `build_index_and_calculate_scores`
- Adds a serializer and the task results of the latest `build_index_and_calculate_scores` run to a new `mtd-data/tasks/` endpoint
- Adds tests for new paths in `build_index_and_calculate_scores` and the new API

### Checklist

### Additional Notes
N/A


firstvoices/backend/views/mtd_data_views.py
```diff
@@ -47,17 +54,37 @@ class MTDSitesDataViewSet(
...
+        site = self.get_validated_site()
+        mtd_exports_for_site = filter_by_viewable(
+            request.user, MTDExportFormat.objects.filter(site=site)
+        )
         if mtd_exports_for_site:
-            return Response(mtd_exports_for_site.latest().latest_export_result)
-        return HttpResponseNotFound(
-            "Site has not been indexed yet. MTD export format not found."
+            return Response(
+                MTDExportFormatTaskSerializer(mtd_exports_for_site.latest()).data
```
bronson-c, line 85: This makes manually confirming that an export_job was cancelled or that an export_job failed difficult as the job instances are deleted after a successful job completes. This shouldn't be a huge issue but if we'd rather keep a log/list of jobs that have been completed I am open to suggestions.

firstvoices/backend/models/mtd.py
```diff
@@ -1,11 +1,11 @@
 from django.db import models
 from django.utils.translation import gettext as _
 
-from backend.models.base import BaseSiteContentModel
+from backend.models.jobs import BaseJob
 from backend.permissions import predicates
 
 
-class MTDExportFormat(BaseSiteContentModel):
+class MTDExportFormat(BaseJob):
```
sahil97, line 8: Since its inheriting from `BaseJob`, would `MTDExportJob` be a more appropriate name now ?

firstvoices/backend/serializers/mtd_serializers.py
```diff
@@ -0,0 +1,23 @@
+from rest_framework import serializers
+
+from backend.models import MTDExportFormat
+from backend.models.jobs import JobStatus
+from backend.serializers import fields
+
+
+class MTDExportFormatTaskSerializer(serializers.ModelSerializer):
```
sahil97, line 8: Similar to above, this then can be `MTDExportJobSerializer`

firstvoices/backend/tests/test_tasks/test_build_mtd.py
```diff
@@ -211,13 +212,49 @@ def test_build_and_score_bad_image_data(self, site, caplog):
...
 
         # Check that only the most recent is in the db
         saved_results = MTDExportFormat.objects.filter(site=site)
         assert len(saved_results) == 1
-        assert saved_results.latest().latest_export_result == final_result
-        assert not saved_results.latest().is_preview
+        assert saved_results.latest().export_result == final_result
+
+    @pytest.mark.django_db
+    def test_build_and_score_not_triggered_while_running(self, site, caplog):
```
sahil97, line 223: One more optional name change suggestion, `test_parallel_build_and_score_jobs_not_allowed` for this test

firstvoices/backend/views/mtd_data_views.py
sahil97: If there is a possibility that we might need to refer back to those records, +1 for keeping a log/list for all the jobs that completed or failed, without storing there export-data possibly ?
But since we can rebuild the mtd-Index as required, and it is not dependent on the previous export_job, I believe we should be good with cleaning out old jobs for now.

sahil97 (approved):
Other than a couple of minor optional name changes, LGTM !

firstvoices/backend/views/mtd_data_views.py
bronson-c: To remedy this I have made the mtd-data/tasks/ endpoint able to serialize multiple models, and multiple models will only exist if there have been multiple failed/cancelled jobs. This means we can see all consecutive mtd export jobs that completed with problems, or just the latest job if the build was successful.

sarahfirstvoices (approved):

PR #1035, opened 2024-09-04
FW-6075 return 401 if user is unauthenticated / cannot view site detail

bronson-c:
### Description of Changes
- Updates FVPermissionViewSetMixin to return `NotAuthenticated` if the user does not have the "view_site" permission
- Ensures `NotAuthenticated` responses always return 401 for sites
- Updates API docs and tests as necessary

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/tests/test_apis/test_sites_api.py
+++ b/firstvoices/backend/tests/test_apis/test_sites_api.py
@@ -592,14 +592,25 @@ class TestSitesEndpoints(MediaTestMixin, BaseApiTest):
         assert response_data["slug"] == str(site.slug)
 
     @pytest.mark.django_db
-    def test_detail_401(self):
+    def test_detail_401_unauthenticated(self):
+        site, user = factories.get_site_with_anonymous_user(
+            visibility=Visibility.MEMBERS
+        )
+        self.client.force_authenticate(user=user)
+
+        response = self.client.get(f"{self.get_detail_endpoint(site.slug)}")
+
+        assert response.status_code == 401
+
+    @pytest.mark.django_db
+    def test_detail_403(self):
         site = factories.SiteFactory.create(visibility=Visibility.MEMBERS)
         user = factories.get_non_member_user()
         self.client.force_authenticate(user=user)
 
         response = self.client.get(f"{self.get_detail_endpoint(site.slug)}")
 
-        assert response.status_code == 401
+        assert response.status_code == 403
 
     @pytest.mark.django_db
     def test_detail_404(self):
@@ -955,7 +966,7 @@ class TestSitesEndpoints(MediaTestMixin, BaseApiTest):
         assert len(response_data["results"]) == 0
 
     @pytest.mark.django_db
-    def test_hidden_site_detail_401(self):
+    def test_hidden_site_detail_403(self):
         user = factories.get_non_member_user()
         self.client.force_authenticate(user=user)
         instance = factories.SiteFactory.create(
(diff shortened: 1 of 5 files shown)


firstvoices/backend/tests/test_apis/test_sites_api.py
```diff
@@ -592,14 +592,14 @@ def test_detail_team_access(self):
...
-    def test_detail_403(self):
+    def test_detail_401(self):
         site = factories.SiteFactory.create(visibility=Visibility.MEMBERS)
         user = factories.get_non_member_user()
         self.client.force_authenticate(user=user)
 
         response = self.client.get(f"{self.get_detail_endpoint(site.slug)}")
 
-        assert response.status_code == 403
+        assert response.status_code == 401
```
sarahfirstvoices, line 602: Ohhhh sorry, this should only return 401 if the user is not logged in. If they lack permissions, 403 is the correct response.

firstvoices/backend/tests/test_apis/test_sites_api.py
```diff
@@ -955,18 +955,18 @@ def test_hidden_sites_visible_in_list_only_to_team(self, team_role):
         assert len(response_data["results"]) == 0
 
     @pytest.mark.django_db
-    def test_hidden_site_detail_403(self):
+    def test_hidden_site_detail_401(self):
```
sarahfirstvoices, line 958: This should apply to all APIs that handle permissions, so please fix the test here as well as in the BaseSiteContentApiTest

firstvoices/backend/views/base_views.py
```diff
@@ -94,6 +95,8 @@ def initial(self, *args, **kwargs):
         # Finally, check permission
         perm = self.get_queryset().model.get_perm(perm_type)
         if not self.request.user.has_perm(perm, obj):
+            if perm == "backend.view_site":
```
sarahfirstvoices, line 98: This condition should instead be something like "if not user.is_authenticated()"

firstvoices/backend/views/sites_views.py
```diff
@@ -32,7 +34,7 @@
         description="Basic information about a language site, for authorized users.",
         responses={
             200: inline_site_doc_detail_serializer,
-            403: OpenApiResponse(description=doc_strings.error_403),
+            401: OpenApiResponse(description=doc_strings.error_401),
```
sarahfirstvoices, line 37: No need to add 401 to the response list here. We list 401 as a possibility for all APIs at the top of the docs, along with 500 and a few others.

firstvoices/backend/views/sites_views.py
```diff
@@ -133,6 +135,18 @@ def get_serializer_context(self):
             context["site"] = self.get_object()
         return context
 
+    def handle_exception(self, exc):
```
sarahfirstvoices, line 138: Is this necessary? I would expect DRF to turn NotAuthenticated into a 401 automatically

sarahfirstvoices (changes requested):

firstvoices/backend/tests/test_apis/test_sites_api.py
bronson-c: Ah apologies I should have asked for clarification on the ticket. I'll adjust the PR to only return a 401 if the user is unauthenticated!

firstvoices/backend/views/base_views.py
```diff
@@ -130,6 +135,18 @@ def list(self, request, *args, **kwargs):
 
         return Response(serializer.data)
 
+    def handle_exception(self, exc):
```
bronson-c, line 138: Currently this block of code seems necessary as the built-in drf framework sets NotAuthenticated calls to 403 when missing an auth header. There is some more drf code setting 401s to 403s as well but it need further investigation Monday.

See `handle_exception` in `/site-packages/rest_framework/views.py` -> `APIView`

bronson-c:
Closing as this ticket would require fixing authentication headers, an entirely different ticket with a much larger scope. Details here: https://firstvoices.atlassian.net/browse/FW-6075

PR #1036, opened 2024-09-06
FW-5604 confirm a batch import bug fixes

sahil97:
### Description of Changes
- Override `import_row` method of `DictionaryEntryResource` to mark invalid and error rows as skipped, thus allowing the valid rows to be imported instead of failing the complete file
- Hardcoded `skipped_rows` in the validation_report as 0 for now

### Checklist

### Additional Notes
N/A


firstvoices/backend/tasks/import_job_tasks.py
```diff
@@ -113,38 +114,24 @@ def import_resource(
         site=import_job_instance.site,
         importjob=import_job_instance,
         new_rows=result.totals["new"],
-        skipped_rows=result.totals["skip"],
-        error_rows=result.totals["error"] + result.totals["invalid"],
+        skipped_rows=0,
```
sahil97, line 117: Do we want to leave a hard-coded variable here ? or we can just remove it from the model, serializer and other places and add it back later when we have a use for it ?

sarahfirstvoices: Let's remove it for now

firstvoices/backend/resources/dictionary.py
```diff
@@ -84,5 +85,32 @@ def before_import(self, dataset, **kwargs):
...
             dataset.append_col(lambda x: str(self.site.id), header="site")
 
+    def import_row(self, row, instance_loader, **kwargs):
+        # Marking erroneous and invalid rows as skipped, then clearing the errors and validation_errors
+        # so the valid rows can be imported
+        import_result = super().import_row(row, instance_loader, **kwargs)
+        if import_result.import_type in [
+            RowResult.IMPORT_TYPE_ERROR,
+            RowResult.IMPORT_TYPE_INVALID,
+        ]:
```
sarahfirstvoices, line 95: Since we always want to skip rows with any type of error, could we invert this condition to something more like `if import_result.import_type != RowResult.SUCCESS` ?

sarahfirstvoices: I don't know how many types of errors or success statuses there are, I just wouldn't want to have an unexpected error type come up and miss this condition

sarahfirstvoices (commented):
Looks good to me as a plan!

firstvoices/backend/resources/dictionary.py
bronson-c: +1!

bronson-c (commented):
As previously discussed this approach LGTM! Will approve when no longer a draft and after Sarah's comments are addressed

firstvoices/backend/resources/dictionary.py
sahil97: Yes, so the plan I had in my mind was just to include these 2 status, `ERROR` and `INVALID` into skipped rows, and not combine all other statuses here. This leaves us room to handle other cases separately in future if required. I understand the concern, but from their `RowResult` enum, these are the only 2 statuses that are returned in case the import doesn't go according to the plan.

The other statuses are `DELETE`, `UPDATE`, `NEW`, `SKIP` for their relevant specific cases.

sarahfirstvoices (approved):

PR #1054, opened 2024-09-16
FW-5096 convert optional charfields in API to non-nullable

bronson-c:
### Description of Changes
- Adds a `ValidateNonNullableCharFieldsMixin` to most writable serializers, which checks for charfields on the model and if they are null and the field is provided, sets them to "" via the validate method
- Changes the optional charfields serializers to allow_null=true, as validation will set them to "" (otherwise the API automatically returns a 400 error)
- Adds the `test_create_with_null_optional_charfields_success_201` and `test_update_with_null_optional_charfields_success_200` tests to most writable APIs
- Skips the above tests if the API does not have eligible optional charfields

### Checklist

### Additional Notes
- Unfortunately a `NonNullableCharfield` solution is not viable, as django raises an integrity error if given a null value for a model field that does not allow null values before the `to_internal_value` method of the field can be applied. The `ValidateNonNullableCharFieldsMixin` performs essentially the same function.

### DIFF
--- a/firstvoices/backend/serializers/song_serializers.py
+++ b/firstvoices/backend/serializers/song_serializers.py
@@ -36,11 +36,13 @@ class SongSerializer(
     visibility = WritableVisibilityField(required=True)
 
     title_translation = serializers.CharField(
-        required=False, allow_blank=True, default=""
+        required=False, allow_blank=True, allow_null=True, default=""
     )
-    introduction = serializers.CharField(required=False, allow_blank=True, default="")
+    introduction = serializers.CharField(
+        required=False, allow_blank=True, allow_null=True, default=""
+    )
     introduction_translation = serializers.CharField(
-        required=False, allow_blank=True, default=""
+        required=False, allow_blank=True, allow_null=True, default=""
     )
 
     lyrics = LyricSerializer(many=True)
--- a/firstvoices/backend/tests/test_apis/test_dictionary_api.py
+++ b/firstvoices/backend/tests/test_apis/test_dictionary_api.py
@@ -358,19 +358,19 @@ class TestDictionaryEndpoint(
             },
         ]
 
-        @pytest.mark.skip(
-            reason="Dictionary entry API does not have eligible optional charfields."
-        )
-        def test_create_with_null_optional_charfields_success_201(self):
-            # Dictionary entry API does not have eligible optional charfields.
-            pass
+    @pytest.mark.skip(
+        reason="Dictionary entry API does not have eligible optional charfields."
+    )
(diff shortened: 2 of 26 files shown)


firstvoices/backend/tests/test_apis/test_widgets_api.py
```diff
@@ -167,6 +164,20 @@ def test_list_minimal(self):
         # Skipping test as it is same as the test_list_permissions test above.
         pass
 
+    @pytest.mark.skip(
+        reason="Site widget API does not have eligible optional charfields."
```
bronson-c, line 168: The only optional charfields on widgets is the "value" in the key/values pairs within the widget settings list, so a null value would just be [] I believe

firstvoices/backend/tests/test_apis/test_dictionary_api.py
```diff
@@ -358,6 +358,20 @@ def assert_update_patch_response(self, original_instance, data, actual_response)
             },
         ]
 
+    @pytest.mark.skip(
```
bronson-c, line 361: The only optional text fields on entries are within arrayFields, so a null value would be [] rather than None which we already handle.

sahil97 (approved):
Looks good, and since the charFields in models does not allow both `blank` and `null`, we should not have any inconsistent data problems in the db as well.

firstvoices/backend/serializers/song_serializers.py
```diff
@@ -36,11 +36,13 @@ class SongSerializer(
...
-        required=False, allow_blank=True, default=""
+        required=False, allow_blank=True, allow_null=True, default=""
+    )
+    introduction = serializers.CharField(
+        required=False, allow_blank=True, allow_null=True, default=""
     )
-    introduction = serializers.CharField(required=False, allow_blank=True, default="")
     introduction_translation = serializers.CharField(
-        required=False, allow_blank=True, default=""
+        required=False, allow_blank=True, allow_null=True, default=""
```
sarahfirstvoices, line 45: What difference does it make to have the default value here? I am wondering whether all these optional CharFields should have the same settings or whether some of them need this extra default value

bronson-c: Oh good point! After a bit of local testing I found that there is no effect on this default value changing `None` to `""`. It seems like it would be used to change `""` values to some other string value if `""` was provided by a PUT request. As per: https://www.django-rest-framework.org/api-guide/fields/#core-arguments

For consistency though I can update the other charfields with `default=""` as that does describe the behaviour that occurs. What do you think?

sahil97 (approved):

PR #1056, opened 2024-09-18
FW-4978 add management command to convert mp3 media models to audio

bronson-c:
### Description of Changes
- adds a management command script that finds all non-audio media models with mp3 files, creates an equivalent audio model with the same data, and then deletes the old mismatched media model
- adds a function to re-link any related models if there is not already an equivalent audio model for the related item
- adds tests for command functionality

### Checklist

### Additional Notes
N/A


firstvoices/backend/management/commands/convert_mp3_to_audio.py
```diff
@@ -0,0 +1,84 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--site",
+            dest="site_slug",
+            help="Site slug of the site to convert media models for (optional)",
+            default=None,
+        )
+
+    @staticmethod
+    def convert_to_audio_model(model):
```
bronson-c, line 23: A possible issue with this conversion is the original ID of the media model is not preserved when converting to an audio model, but all other data is. If this will be a problem I believe I can save the ID and change it after the deleting on the original model

sarahfirstvoices: We should check whether any of the affected non-Audio mp3s are related to anything else. If they have no relationships, it's fine to reset the IDs. If they are linked, we need to think a little more about how to preserve those links. I am guessing there are only going to be a couple of cases and we can problem re-link them by hand

bronson-c: After doing some research on preprod data there are 39 mp3 images and 10 mp3 videos with related content (characters, dictionary entries, songs and stories). I will look into a solution that keeps the original ids, or at least can preserve the relationship between related media

sarahfirstvoices: Annoyingly, I suppose we should also check those 39+10 items to see whether the same mp3 was already also added as audio

sahil97 (approved):
LGTM !

PR #1057, opened 2024-09-20
FW-5605 Ability to get a csv for the failed rows while batch importing

sahil97:
### Description of Changes
- Added field to import-job model for failed_rows csv file,
- added generation and attaching this csv file in the batch import task
- added relevant tests

### Checklist
- ~ README / documentation has been updated if needed~

### Additional Notes
N/A


sahil97:
Do we want to add a ticket to possibly add a cleanup task or something similar for the batch import files ? After this addition, we will be keeping atleast 2 CSVs per batch. Would that be of concern in context of storage and/or cost ?

firstvoices/backend/tests/test_tasks/test_import_job_tasks.py
```diff
@@ -322,6 +329,39 @@ def test_dry_run_failed(self, caplog):
...
+            format="csv",
+        )
+
+        assert len(failed_rows_csv_table) == 5
+
+        for i in range(0, len(error_rows_numbers)):
+            input_index = (
+                error_rows_numbers[i] - 1
+            )  # since we do +1 while generating error row numbers
+            assert input_csv_table[input_index] == failed_rows_csv_table[i]
```
sahil97, line 363: This is verifying each row that is returned is the same as the row in the input file without any modifications from us.

firstvoices/backend/models/import_jobs.py
```diff
@@ -116,3 +116,10 @@ class Meta:
     validation_report = models.OneToOneField(
         ImportJobReport, null=True, on_delete=models.SET_NULL
     )
+
+    failed_rows_csv = models.ForeignKey(
+        File,
+        null=True,
+        on_delete=models.SET_NULL,
+        related_name="import_job_failed_rows_csv_set",
+    )
```
sarahfirstvoices, line 125: Could this be one-to-one instead of foreign key? It makes some things easier

firstvoices/backend/tasks/import_job_tasks.py
```diff
@@ -89,12 +96,42 @@ def clean_csv(data):
...
+        name="failed_rows.csv",
+        content_type="text/csv",
+        size=sys.getsizeof(in_memory_csv_file),
+        charset="utf-8",
+    )
+    failed_row_csv_file = File(
+        content=in_memory_csv_file,
+        site=import_job_instance.site,
+        created_by=import_job_instance.created_by,
+        last_modified_by=import_job_instance.last_modified_by,
```
sarahfirstvoices, line 131: Totally minor, but I think the created_by and last_modified_by here should both be set to the import_job_instance.last_modified_by value just in case there is any difference

sarahfirstvoices (changes requested):
Just minor change suggestions. Looking good otherwise!

firstvoices/backend/models/import_jobs.py
sahil97: Kept as ForeignKey but updated the data field above in the same model to also be a ForeignKey as discussed.

sarahfirstvoices (approved):

bronson-c (approved):

PR #1067, opened 2024-09-25
FW-6156 Add media sync to admin site operations

bronson-c:
### Description of Changes
- Currently site media was only syncing if a site feature was changed in the API, this PR adds the sync to admin site operations as well

### Checklist
- [x] ~~Tests have been updated / created if applicable~~

### Additional Notes
- Unsure if unit tests for this change were necessary, as the admin site is only for staff use and does not have any prior unit tests.

### DIFF
--- a/firstvoices/backend/admin/sites_admin.py
+++ b/firstvoices/backend/admin/sites_admin.py
@@ -8,6 +8,9 @@ from backend.models.sites import (
     SiteMenu,
 )
 
+from ..search.tasks.site_content_indexing_tasks import (
+    request_sync_all_media_site_content_in_indexes,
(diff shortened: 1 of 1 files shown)


sahil97 (approved):
Looks good!
Just wondering if we should add this update to a signal triggered by SiteFeature instead ? we can add a check to ensure that we are only syncing the media in case of the right SiteFeature change.
Not completely sure if that will work, but I believe it can help keep most of the code handling index updates consolidated in one place.

sarahfirstvoices (approved):
Good call doing it in the admin form and not right on the model for this one weird case.

PR #1084, opened 2024-10-08
Sonar coverage fixes

sahil97:
### Description of Changes
- Added back source to coveragerc

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/.coveragerc
+++ b/firstvoices/.coveragerc
@@ -1,5 +1,5 @@
 [run]
-source=['firstvoices']
+source=['.', './firstvoices']
 branch=True
 omit =
     */tests/*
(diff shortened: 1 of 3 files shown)


bronson-c (approved):
LGTM 👍

firstvoices/.coveragerc
```diff
@@ -1,13 +1,6 @@
...
-    firstvoices/*
-    healthcheck/*
-    jwt_auth/*
-    scripts/*
-    manage.py
+source=.
+branch=True
 omit =
     */tests/*
     */migrations/*
```
sarahfirstvoices, line 6: Looks like this is still excluding the resources folder from back when we were focused on the initial data migrations. Now that we're using the resources for application code I think they should get included.

Not sure why the search tasks are also excluded-- can they be included as well or is there some problem?

I think the fixtures folder could be excluded though. It probably gets loaded by every test, but it isn't meaningful lines of code and will skew our percentages.

PR #1123, opened 2024-11-15
Fw 6275 add validate endpoint for an import job

sahil97:
### Description of Changes
- Added `import-jobs/{import-job-id}/validate` endpoint to validate an import-job
- Made importJob optional in `File` model to fix admin interface requiring importJob for media files
- Checks performed before starting a validation job
  - No other job on the same site is queued or running, either for validation or db import
  - The specified job's dry run is not queued or running. 

### Checklist

### Additional Notes
N/A


firstvoices/backend/tests/test_apis/test_import_job_api.py
```diff
@@ -395,21 +437,90 @@ def test_confirm_action_404_unknown_key(self):
...
             content_type=self.content_type,
         )
+        assert response.status_code == 201
         response_data = json.loads(response.content)
 
-        assert "failedRowsCsv" in response_data
+        self.import_job = ImportJob.objects.filter(id=response_data["id"])[0]
+
+    @pytest.mark.django_db(transaction=True)
+    def test_exception_fetching_previous_report(self, caplog):
```
bronson-c, line 488: [minor] I believe you should also assert that the self.import_job.validation_status = JobStatus.FAILED in this test

bronson-c (dismissed):
LGTM! Just a small nitpick with one of the tests

bronson-c (dismissed):
LGTM!

firstvoices/backend/tasks/import_job_tasks.py
```diff
@@ -211,6 +211,17 @@ def import_job_dry_run(data, import_job_instance, logger):
         import_job=import_job_instance.id,
     )
 
+    # Clearing out old report if revalidate
+    old_report = import_job_instance.validation_report
```
sarahfirstvoices, line 215: What if there is an old report but they don't choose revalidate? Are we intentionally leaving the old report?

firstvoices/backend/views/import_job_views.py
```diff
@@ -159,3 +185,24 @@ def confirm(self, request, site_slug=None, pk=None):
         return Response(
             serializer.data, status=status.HTTP_202_ACCEPTED, headers=headers
         )
+
+    @action(detail=True, methods=["post"])
+    def validate(self, request, site_slug=None, pk=None):
+        """
+        Method to start the validation process on a given import-job.
+        """
+        import_job_id = self.kwargs["pk"]
```
sarahfirstvoices, line 194: If there is already a validation in progress from the initial creation, what happens? Should we return a 400 error at this point? We don't have a way to cancel the first validation, and there is no guarantee the two reports will complete in the right order. I think we need to think through the collision cases a bit more

sarahfirstvoices (changes requested):
Let's chat about the async aspects a bit and make sure all the paths will work

firstvoices/backend/views/import_job_views.py
sahil97: With the changes I just added, if the `/validate` endpoint is hit while a job is queued or running (same job or a different job), it will give a 400 and state that the user should wait for the current job to finish.

firstvoices/backend/tasks/import_job_tasks.py
```diff
@@ -265,7 +290,7 @@ def batch_import(import_job_instance_id, dry_run=True):
         logger.info(ASYNC_TASK_END_TEMPLATE)
         return
 
-    # If dry-run has not been executed successfully, do not proceed
+    # If dry-run has not been executed successfully, do not proceed for the db import
```
sarahfirstvoices, line 293: Unrelated to these changes, but I just noticed the error log on line 297. Error logs are meant for situations where the code is not running as expected and we need an admin to go fix something asap. This is just a user error being handled appropriately-- I think it should be an info log instead, to avoid noisy alerts for us

sarahfirstvoices: That log message actually looks like it is intended for the user and not for the logs-- should we put it into the message on the job instead? Or possibly move this check into the API so we could return a 400 status in this case?

firstvoices/backend/views/import_job_views.py
```diff
@@ -159,3 +193,49 @@ def confirm(self, request, site_slug=None, pk=None):
...
+        logger = logging.getLogger(__name__)
+
+        # Verify that no other jobs are started or queued for the same site
+        existing_incomplete_jobs = get_import_jobs_queued_or_running(
+            site, import_job_id
+        )
+
+        if len(existing_incomplete_jobs):
+            logger.error(
```
sarahfirstvoices, line 212: If this message is meant for the user, it needs to go in the Response object instead of into the logs.

firstvoices/backend/views/import_job_views.py
```diff
@@ -159,3 +193,49 @@ def confirm(self, request, site_slug=None, pk=None):
...
+            )
+            logger.info(ASYNC_TASK_END_TEMPLATE)
+            return Response(status=status.HTTP_400_BAD_REQUEST)
+
+        # Edge case check to prevent running validation on the current job
+        # if its already running or queued for dry run
+        curr_job = ImportJob.objects.filter(id=import_job_id)[0]
+        if curr_job.validation_status in [JobStatus.ACCEPTED, JobStatus.STARTED]:
+            logger.error(
+                "The specified job is already running or queued. "
```
sarahfirstvoices, line 224: Same issue here-- user messages need to go in the Response object. The easiest way is usually to raise a ValidationError rather than manually constructing the response. Something like `raise raise ValidationError("The specified job is already running etc etc...")`

sarahfirstvoices (changes requested):
I added a few more comments about logging and error responses, but the async fixes look good to me!

sarahfirstvoices (approved):

PR #1129, opened 2024-11-19
FW-2071 Enhance cross-site search API to search by list of sites

bronson-c:
### Description of Changes
- Adds the site slug keyword to site content documents and indexes
- Adds a "sites" query parameter that exact matches on site slug, and can take multiple slugs separated by comma
- If any slug is valid, returns the data from valid slugs
- If all slugs are invalid, returns an empty search list

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/search/query_builder.py
+++ b/firstvoices/backend/search/query_builder.py
@@ -18,7 +18,6 @@ from backend.search.utils.query_builder_utils import (
     get_kids_query,
     get_max_words_query,
     get_min_words_query,
-    get_multi_site_filter_query,
     get_site_filter_query,
     get_starts_with_query,
     get_types_query,
@@ -36,7 +35,6 @@ def get_search_object(indices):
 def get_search_query(
     user=None,
     q=None,
-    site_id=None,
     sites=None,
     types=VALID_DOCUMENT_TYPES,
     domain="both",
@@ -88,11 +86,8 @@ def get_search_query(
             )
 
     # Add site filter if parameter provided in url
-    if site_id:
-        search_query = search_query.query(get_site_filter_query(site_id))
-
     if sites:
-        search_query = search_query.query(get_multi_site_filter_query(sites))
+        search_query = search_query.query(get_site_filter_query(sites))
 
     types_query = get_types_query(types)
     if types_query:
@@ -100,7 +95,7 @@ def get_search_query(
 
     if starts_with_char:
         search_query = search_query.query(
-            get_starts_with_query(site_id, starts_with_char)
+            get_starts_with_query(sites[0], starts_with_char)
         )
 
     if category_id:
--- a/firstvoices/backend/search/query_builder.py
+++ b/firstvoices/backend/search/query_builder.py
@@ -18,8 +18,8 @@ from backend.search.utils.query_builder_utils import (
     get_kids_query,
     get_max_words_query,
     get_min_words_query,
+    get_multi_site_filter_query,
     get_site_filter_query,
-    get_site_slug_filter_query,
     get_starts_with_query,
     get_types_query,
     get_view_permissions_filter,
@@ -92,7 +92,7 @@ def get_search_query(
         search_query = search_query.query(get_site_filter_query(site_id))
 
     if sites:
-        search_query = search_query.query(get_site_slug_filter_query(sites))
+        search_query = search_query.query(get_multi_site_filter_query(sites))
 
     types_query = get_types_query(types)
     if types_query:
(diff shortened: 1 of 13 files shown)


firstvoices/backend/search/utils/query_builder_utils.py
```diff
@@ -80,6 +80,10 @@ def get_site_filter_query(site_id):
     return Q("bool", filter=[Q("term", site_id=site_id)])
 
 
+def get_site_slug_filter_query(site_slug):
+    return Q("bool", filter=[Q("terms", site_slug=site_slug)])
```
sarahfirstvoices, line 84: The param should be plural here-- site_slugs

firstvoices/backend/search/utils/validators.py
```diff
@@ -148,3 +150,23 @@ def get_valid_site_feature(input_site_feature_str):
     if len(selected_values) == 0:
         return None
     return selected_values
+
+
+def get_valid_site_slugs(input_site_slug_str, user):
```
sarahfirstvoices, line 155: Since we are already making a db query here, why not return the valid site IDs instead of the slugs? It would avoid needing to index the site slug in every index. This function could be get_valid_site_ids_from_slugs

firstvoices/backend/search/utils/validators.py
```diff
@@ -148,3 +150,23 @@ def get_valid_site_feature(input_site_feature_str):
...
+    input_site_slugs = input_site_slug_str.split(",")
+    valid_site_slugs = filter_by_viewable(user, Site.objects.all()).values_list(
+        "slug", flat=True
+    )
+    selected_values = []
+
+    for value in input_site_slugs:
+        cleaned_value = value.strip().lower()
+        if cleaned_value in valid_site_slugs and cleaned_value not in selected_values:
+            selected_values.append(cleaned_value)
```
sarahfirstvoices, line 168: A simpler way to do this might be:

input_site_slugs = [value.strip().lower() for value in input_site_slug_str.split(",")]
selected_values = Site.objects().visible(user).filter(slug__in=input_site_slugs)

firstvoices/backend/views/search_all_entries_views.py
```diff
@@ -516,6 +517,25 @@
...
+                    OpenApiExample(
+                        "",
+                        value="",
+                        description="Default case. Do not add sites filter.",
+                    ),
+                    OpenApiExample(
+                        "site1, site2",
+                        value="site1, site2",
+                        description="Return entries which are associated with the specified sites.",
+                    ),
```
sarahfirstvoices, line 536: Good docs!

firstvoices/backend/views/search_all_entries_views.py
```diff
@@ -658,6 +682,9 @@ def list(self, request, **kwargs):
         if search_params["visibility"] is None:
             return self.paginate_search_response(request, [], 0)
 
+        if search_params["sites"] is None:
+            return self.paginate_search_response(request, [], 0)
```
sarahfirstvoices, line 686: Does this mean that `?sites=` will return no results?

sarahfirstvoices (changes requested):

firstvoices/backend/views/search_all_entries_views.py
bronson-c: `search_params["sites"]` only returns None if all site slug are invalid, otherwise it returns "".  `?sites=` currently would return all entries, the same as how the types and visibility params work

firstvoices/backend/views/search_all_entries_views.py
```diff
@@ -574,13 +594,17 @@ def get_search_params(self):
...
+
         search_params = {
             "q": input_q,
             "user": user,
             "types": valid_types_list,
             "domain": valid_domain,
             "kids": kids_flag,
             "games": games_flag,
+            "sites": valid_site_ids,
             "site_id": "",  # used in site-search
```
sarahfirstvoices, line 608: I'm looking at these two parameters-- sites and site_id. What if we combine them? The rest APIs would stay the same, but this function could just use the sites list. For the site-specific search, we could put the single site_id into the sites list parameter. It would avoid any weird scenarios where we accidentally try to use both at the same time

sarahfirstvoices (changes requested):

firstvoices/backend/search/query_builder.py
```diff
@@ -86,16 +86,16 @@ def get_search_query(
...
 
     types_query = get_types_query(types)
     if types_query:
         search_query = search_query.query(get_types_query(types))
 
     if starts_with_char:
         search_query = search_query.query(
-            get_starts_with_query(site_id, starts_with_char)
+            get_starts_with_query(sites[0], starts_with_char)
         )
```
sarahfirstvoices, line 99: Oh interesting combo case.... I'm going to log a ticket to design how this should actually work. This will work for now but it will give unexpected results if people try to use starts_with_char in the cross-site search

sarahfirstvoices (approved):

PR #1168, opened 2025-01-14
FW-6287 Handle case-insensitive inputs for PartOfSpeech in DictionaryEntryResource

sahil97:
### Description of Changes
- Handle case-insensitive matches for PartOfSpeech in DictionaryEntryResource

### Checklist

### Additional Notes
N/A


firstvoices/backend/resources/utils/import_export_widgets.py
```diff
@@ -117,18 +120,13 @@ def clean(self, value, row=None, **kwargs):
         ]
 
 
-class CleanForeignKeyWidget(ForeignKeyWidget):
-    def __init__(self, model, field, title_case=False, *args, **kwargs):
-        self.title_case = title_case
-        super().__init__(model=model, field=field, *args, **kwargs)
+class PartOfSpeechWidget(ForeignKeyWidget):
```
sahil97, line 123: Can be later renamed again as `CleanForeignKeyWidget` and made reusable if required, but was currently only used here so I put the PartOfSpeech in the name.

firstvoices/backend/resources/utils/import_export_widgets.py
```diff
@@ -117,18 +120,13 @@ def clean(self, value, row=None, **kwargs):
...
+        super().__init__(model=PartOfSpeech, field="title", *args, **kwargs)
 
     def clean(self, value, row=None, **kwargs):
         value = value.strip().lower()
-
-        if self.title_case:
-            value = value.title()
-
-        return super().clean(value, row, **kwargs)
+        return get_object_or_404(PartOfSpeech, title__iexact=value)
```
sarahfirstvoices, line 129: Do we need any of the things from the parent's clean method?

sahil97: I don't believe so. The clean method in the library is strangely not doing any cleaning, but just tries to return values based on `natural_key`/`primary_key` or any other lookup_kwargs we provide. 
Since we are returning a list of objects directly, I believe that we can safely skip that.

bronson-c (approved):
LGTM!

PR #1190, opened 2025-02-06
FW-6350 documents apis

gmcauliffe:
### Description of Changes

- Added Documents APIs and tests

### Checklist

### Additional Notes
N/A


firstvoices/backend/serializers/media_detail_serializers.py
```diff
@@ -95,6 +96,11 @@ class Meta(AudioSerializer.Meta):
         fields = AudioSerializer.Meta.fields + BaseUsageFieldSerializer.Meta.fields
 
 
+class DocumentDetailSerializer(DocumentSerializer):
+    class Meta(DocumentSerializer.Meta):
+        fields = DocumentSerializer.Meta.fields
```
sarahfirstvoices, line 101: If there is no change from the parent serializer, there is no need to add this extra serializer. BUT, I think we should keep this and add the fields from the BaseUsageFieldSerializer

sahil97 (approved):
LGTM !

firstvoices/backend/serializers/media_detail_serializers.py
gmcauliffe: For sure. Will add the BaseUsageFieldSerializer with the related_documents ticket 👍

sarahfirstvoices: Which I realize will only work once you do the related ticket... so this is good as-is for now :)

firstvoices/backend/tests/test_apis/base_media_test.py
```diff
@@ -49,16 +50,35 @@ def get_basic_media_data(self, instance, view_name, detail_view):
...
+
+    def get_image_usage_data(self):
+        usage = {
+            **self.get_video_usage_data(),
+            "gallery": [],
+            "siteBanner": {},
+            "siteLogo": {},
+        }
+
+        return usage
```
sarahfirstvoices, line 80: Great update for these usage fields tests

firstvoices/backend/tests/test_apis/base_media_test.py
```diff
@@ -687,6 +723,8 @@ def test_usages_field_base(self):
 
     @pytest.mark.django_db
     def test_usages_field_permissions(self):
+        if self.model == Document:
+            pytest.skip("Documents don't currently have usages.")
```
sarahfirstvoices, line 727: No need to change this because I know it's very temporary, but for future reference I recommend skipping tests in the child test class instead, like this: https://github.com/First-Peoples-Cultural-Council/fv-be/blob/a905565821c2d121da6b17d2a1a628cb977610b5/firstvoices/backend/tests/test_apis/test_bulkvisibility_api.py#L75

It has better visibility (you can look at the document test class and see what it runs and what it skips), and it avoids having to check for interactions in all the other child classes.

gmcauliffe: I was wondering what the best way to do this was. Thanks for the tip. Much appreciated!

PR #1191, opened 2025-02-07
FW-6293 Added status check for import-job delete action

sahil97:
### Description of Changes
- Added status check for import-job delete action.

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/models/import_jobs.py
+++ b/firstvoices/backend/models/import_jobs.py
@@ -1,5 +1,6 @@
 from django.contrib.auth import get_user_model
 from django.contrib.postgres.fields import ArrayField
+from django.core.exceptions import ValidationError
 from django.db import models
 from django.utils.translation import gettext as _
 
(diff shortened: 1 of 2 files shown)


bronson-c (dismissed):
LGTM!

firstvoices/backend/models/import_jobs.py
```diff
@@ -121,3 +122,18 @@ class Meta:
...
+    def _delete_report(self):
+        import_job_report = self.validation_report
+        if import_job_report:
+            import_job_report.delete()
+
+    def delete(self, using=None, keep_parents=False):
+        """
+        Does not allow deleting on an instance if the job has been completed, i.e. status="completed".
+        """
+        if self.status == JobStatus.COMPLETE:
```
sarahfirstvoices, line 135: I think need to prevent deletion in some more cases as well:

- status = STARTED or ACCEPTED, since the task will still run if we delete the job, and we'll have confusing results
- validation_status = STARTED, because the running task will re-save the job and the validation report. In most cases validation_status = ACCEPTED would be ok, since when the task runs it gets the ImportJob early on and would fail, but with async things it is still possible that in between us checking the status and deleting the job, the task could start and get past that check. SO, we should also block deleting a job if validation_status is ACCEPTED

sarahfirstvoices (changes requested):

bronson-c (approved):

sarahfirstvoices (approved):

PR #1209, opened 2025-02-27
FW-6225 Add reset-last-modified flag to copy_site command

sahil97:
### Description of Changes

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/management/commands/copy_site.py
+++ b/firstvoices/backend/management/commands/copy_site.py
@@ -56,16 +56,16 @@ def create_new_site(source_site, target_site_slug, user):
     return new_site
 
 
-def copy_site_features(source_site, target_site):
+def copy_site_features(source_site, target_site, set_modified_date):
     site_features = list(SiteFeature.objects.filter(site=source_site))
     for site_feature in site_features:
         site_feature.id = uuid.uuid4()
         site_feature.site = target_site
-        site_feature.save()
+        site_feature.save(set_modified_date=set_modified_date)
 
 
 def copy_all_characters_and_return_map(
-    source_site, target_site, audio_map, image_map, video_map
+    source_site, target_site, audio_map, image_map, video_map, set_modified_date
 ):
     character_map = {}
     characters = list(Character.objects.filter(site=source_site))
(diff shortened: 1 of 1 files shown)


firstvoices/backend/management/commands/copy_site.py
```diff
@@ -411,32 +399,53 @@ def copy_dictionary_entries_and_return_map(
...
             if source_related_entry.id in dictionary_entry_map
         ]
-        entry.related_dictionary_entries.set(updated_related_entries)
+        target_entry.related_dictionary_entries.set(updated_related_entries)
 
-        copy_related_media(entry, source_media, audio_map, image_map, video_map)
-        entry.save()
+        source_media = {
+            "audio": list(source_entry.related_audio.values_list("id", flat=True)),
+            "images": list(source_entry.related_images.values_list("id", flat=True)),
```
gmcauliffe, line 432: Obviously there aren't any `"documents"`/`related_documents` to copy yet but you may want to add it in here anyway.

gmcauliffe (dismissed):
LGTM. Just one suggestion about including `related_documents` but they could be added at a later stage.

sarahfirstvoices:
I still see a few calls to save() without the parameter. I think that's fine for the files, the site itself, and the gallery items, but it seems like lines 181, 221, and 253 should have the set_modified_date param. What do you think?

sahil97:
@sarahfirstvoices good catch. I skipped adding the save parameter to those calls since we're setting the `_state.adding` just before the save call. The base class save() method conditional would short-circuit and would never reach the check for `set_modified_date` parameter.

`
def save(self, set_modified_date=True, *args, **kwargs):
    """Update last_modified time if updating the model."""
    if (not self._state.adding) and set_modified_date:
        self.last_modified = timezone.now()
    return super().save(*args, **kwargs)
`

firstvoices/backend/management/commands/copy_site.py
```diff
@@ -175,9 +176,9 @@ def copy_audio_and_speakers_and_return_map(source_site, target_site, logger):
             audio_map[source_audio_id] = target_audio_id
 
             audio.id = target_audio_id
-            # To circumvent checks added to media models to prevent modification of original file
+            # To circumvent certain conditionals that were originally added to prevent modification of the original file
```
sarahfirstvoices, line 179: Minor: isn't this actually here to make sure it doesn't get confused when checking whether we need to generate thumbnails?

sarahfirstvoices (dismissed):

firstvoices/backend/management/commands/copy_site.py
sahil97: yep, my bad. Confused this with the file model check. Will update it

gmcauliffe (approved):

PR #1272, opened 2025-04-17
FW-6513 Image and Video upload workflow for import-jobs

sahil97:
### Description of Changes
- Added image and video resources
- Added methods to support importing images and videos
- Added method to ignore duplicate rows with same media filenames, and refer to the first occurrence of that filename for related_media.

### Checklist
- ~Insomnia workspace has been updated if applicable

### Additional Notes
N/A


firstvoices/backend/resources/media.py
```diff
@@ -45,10 +45,10 @@ class AudioResource(ControlledSiteContentResource):
         ),
     )
     exclude_from_kids = fields.Field(
-        column_name="audio_include_on_kids_site",
+        column_name="audio_include_in_kids_site",
         attribute="exclude_from_kids",
         widget=InvertedBooleanFieldWidget(
-            column="audio_include_on_kids_site", default=False
+            column="audio_include_in_kids_site", default=False
```
sarahfirstvoices, line 51: The KB page specifies "on" -- was this changed intentionally?

firstvoices/backend/resources/media.py
```diff
@@ -66,3 +66,87 @@ def before_import_row(self, row, **kwargs):
     class Meta:
         model = Audio
         clean_model_instances = True
+
+
+class ImageResource(ControlledSiteContentResource):
```
sarahfirstvoices, line 71: I realize the AudioResource was already using this base class, but media models are actually UncontrolledSiteContent (no visibility field) so this should get changed.

firstvoices/backend/tasks/import_job_tasks.py
```diff
@@ -215,6 +218,72 @@ def import_audio_resource(import_job, audio_data, dictionary_entry_data, dry_run
...
+    # Adding image ids
+    if img_import_result.totals["new"]:
+        img_lookup = {
+            row["img_filename"]: row["id"]
+            for row in img_data.dict
+            if row.get("img_filename")
+        }
+        dictionary_entry_data.append_col(
+            [""] * len(dictionary_entry_data), header="related_images"
+        )
```
sarahfirstvoices, line 242: Immutable data is always easier to deal with than having a function modify passed-in data in place. (Easier maintenance, fewer mystery bugs.) Instead of passing in and modifying the dictionary_entry_data, could we have these media import methods return dictionaries for looking up the ID by filename? Then we could pass those to the dictionary entry import method? I haven't looked at this as much as you, so I am probably missing some detail. Happy to chat about it on a call.

firstvoices/backend/tasks/import_job_tasks.py
```diff
@@ -215,6 +218,72 @@ def import_audio_resource(import_job, audio_data, dictionary_entry_data, dry_run
...
+        related_video_col_index = dictionary_entry_data.headers.index("related_videos")
+        for i, row in enumerate(dictionary_entry_data.dict):
+            video_filename = row.get("video_filename")
+            related_id = video_lookup.get(video_filename, "")
+            row_list = list(dictionary_entry_data[i])
+            row_list[related_video_col_index] = related_id  # comma separated string
+            dictionary_entry_data[i] = tuple(row_list)
+
+    return video_import_result, dictionary_entry_data
```
sarahfirstvoices, line 284: Totally optional, but I know you were looking for ways to split up or organize this long file. The media import methods share a lot of logic but use different resources. They also depend on similar logic to "split" the csv data based on column names.

One option would be to move all that logic into a set of Importer classes that reuse the same methods and have internal data members to specify the column prefix, resource, etc. Then we could call the actual import methods similarly to:

`image_import_result, image_lookup = ImageImporter.import(import_job, csv, dry_run)`

sarahfirstvoices: The could potentially also keep track of their own valid column header options, and maybe we could find a way to use those for the up-front validation of the csv file as well.

firstvoices/backend/resources/media.py
sahil97: Yes, its slightly confusing. As per this [KB Page](https://firstvoices.atlassian.net/wiki/spaces/FIR1/pages/17664196/Prepare+batch+upload+spreadsheets+for+words+and+phrases#Example-batches)
In the dictionary entries,  its `INCLUDE_ON_KIDS_SITE` and `INCLUDE_IN_GAMES`.
while for media its `AUDIO_INCLUDE_IN_KIDS_SITE` and so on.

firstvoices/backend/resources/media.py
sahil97: Good catch.

firstvoices/backend/resources/media.py
bronson-c: +1

firstvoices/backend/resources/media.py
```diff
@@ -66,3 +66,87 @@ def before_import_row(self, row, **kwargs):
...
+    )
+
+    description = fields.Field(column_name="img_description", attribute="description")
+
+    acknowledgement = fields.Field(
+        column_name="img_acknowledgement", attribute="acknowledgement"
+    )
+
+    exclude_from_kids = fields.Field(
+        column_name="img_include_in_kids_site",
```
bronson-c, line 90: Follow up to Sarah's comment: whether it's "on_kids_site" or "in_kids_site" should match for other media types as well

firstvoices/backend/tasks/import_job_tasks.py
sahil97: That makes sense and I agree on the improvement in maintenance. I'll refactor to add in these changes.

firstvoices/backend/resources/media.py
sarahfirstvoices: Oh that's annoying. I logged a ticket to make them all match, but this is good for now. Thanks for explaining! https://firstvoices.atlassian.net/browse/FW-6546

firstvoices/backend/tasks/import_job_tasks.py
```diff
@@ -263,31 +241,45 @@ def import_video_resource(import_job, video_data, dictionary_entry_data, dry_run
...
+                [""] * len(dictionary_entry_data), header=column_name
+            )
+            col_index = dictionary_entry_data.headers.index(column_name)
+            for i, row in enumerate(dictionary_entry_data.dict):
+                filename = row.get(filename_key)
+                related_id = media_map.get(filename, "")
+                row_list = list(dictionary_entry_data[i])
+                row_list[col_index] = related_id
+                dictionary_entry_data[i] = tuple(row_list)
+
```
sarahfirstvoices, line 282: This works, but it loops through the list of dictionary entries 3 times. It would be more efficient to loop through one time, and look up the three filenames as needed.

sarahfirstvoices (changes requested):
Looking good! Just the one suggestion for performance reasons

firstvoices/backend/tasks/import_job_tasks.py
sahil97: Good idea, I'll try it and see if I can further optimize it.

PR #1326, opened 2025-06-02
FW-6499 ETag and Last-Modified headers for mtd api

gmcauliffe:
### Description of Changes

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/tests/test_apis/test_mtd_data_api.py
+++ b/firstvoices/backend/tests/test_apis/test_mtd_data_api.py
@@ -1,10 +1,13 @@
+import hashlib
+from datetime import datetime
+
 import pytest
 from mothertongues.config.models import LanguageConfiguration
 from rest_framework.reverse import reverse
 from rest_framework.test import APIClient
 
 from backend.models import MTDExportJob
-from backend.models.constants import Visibility
+from backend.models.constants import Role, Visibility
 from backend.models.jobs import JobStatus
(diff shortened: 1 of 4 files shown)


gmcauliffe:
Curious to know people's thoughts on extending use of ETag and Last-Modified headers to other detail apis. 
Strategy/Approach suggestions for code organization?

firstvoices/backend/tests/test_apis/test_mtd_data_api.py
```diff
@@ -129,3 +132,26 @@ def test_data_integrity_build_and_score_failed(self, status):
         )
         assert response.status_code == 200
         assert response.data[0]["status"] == status
+
+    @pytest.mark.django_db
+    def test_etag_and_last_modified(self):
+        self.user = factories.get_superadmin()
+        self.client.force_authenticate(user=self.user)
```
sarahfirstvoices, line 139: Minor: I don't think this request needs to be authenticated at all for testing, since it's available to guests

sarahfirstvoices:
> Curious to know people's thoughts on extending use of ETag and Last-Modified headers to other detail apis. Strategy/Approach suggestions for code organization?

We talked about this in real life but I propose we leave this for a separate ticket. It would take some research to determine which fields and related models we'd need to check for each view, and then some thinking about the best place to put that logic. We could also research which views would be highest value for this type of caching-- maybe the site homepages?

PR #1358, opened 2025-06-26
FW-4337 membership update delete

gmcauliffe:
### Description of Changes
Adds update and delete apis for site Memberships

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/serializers/fields.py
+++ b/firstvoices/backend/serializers/fields.py
@@ -95,7 +95,7 @@ class WritableVisibilityField(serializers.CharField):
 class WritableRoleField(serializers.CharField):
     def to_internal_value(self, data):
         try:
-            return Role[data.upper()]
+            return Role[data.upper().replace(" ", "_")]
         except KeyError:
             raise serializers.ValidationError("Invalid role option.")
 
--- a/firstvoices/backend/serializers/fields.py
+++ b/firstvoices/backend/serializers/fields.py
@@ -5,7 +5,7 @@ from rest_framework.exceptions import ValidationError
 from rest_framework.reverse import reverse
 from rest_framework_nested.relations import NestedHyperlinkedRelatedField
 
-from backend.models.constants import Visibility
+from backend.models.constants import Role, Visibility
 from backend.serializers.utils.context_utils import get_site_from_context
 
 
@@ -90,6 +90,21 @@ class WritableVisibilityField(serializers.CharField):
             return visibility_map[value].lower()
         except KeyError:
             raise serializers.ValidationError("Invalid visibility value.")
+
+
+class WritableRoleField(serializers.CharField):
+    def to_internal_value(self, data):
+        try:
+            return Role[data.upper()]
+        except KeyError:
+            raise serializers.ValidationError("Invalid role option.")
+
+    def to_representation(self, value):
+        role_map = {choice[0]: choice[1] for choice in Role.choices}
+        try:
+            return role_map[value]
+        except KeyError:
+            raise serializers.ValidationError("Invalid role value.")
 
 
 class EnumField(serializers.Field):
--- a/firstvoices/backend/serializers/fields.py
+++ b/firstvoices/backend/serializers/fields.py
@@ -5,7 +5,7 @@ from rest_framework.exceptions import ValidationError
 from rest_framework.reverse import reverse
 from rest_framework_nested.relations import NestedHyperlinkedRelatedField
 
-from backend.models.constants import Role, Visibility
+from backend.models.constants import Visibility
 from backend.serializers.utils.context_utils import get_site_from_context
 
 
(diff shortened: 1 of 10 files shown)


firstvoices/backend/serializers/fields.py
```diff
@@ -92,6 +92,21 @@ def to_representation(self, value):
             raise serializers.ValidationError("Invalid visibility value.")
 
 
+class WritableRoleField(serializers.CharField):
```
sarahfirstvoices, line 95: I think you should be able to use EnumField (directly below) instead of creating a new field for this. That's what we use for Visibility, which is the same kind of data as Role.

firstvoices/backend/serializers/membership_serializers.py
```diff
@@ -1,3 +1,4 @@
+from jwt_auth.models import User
```
sarahfirstvoices, line 1: This is a little Django gotcha. Instead of directly importing the User model, please use the get_user_model() method

https://docs.djangoproject.com/en/4.0/topics/auth/customizing/#referencing-the-user-model

firstvoices/backend/serializers/membership_serializers.py
```diff
@@ -54,19 +56,37 @@ class Meta:
...
-        view_name="api:membership-detail", read_only=True
+        read_only=True, view_name="api:membership-detail"
     )
-    user = UserDetailSerializer(allow_null=False, read_only=True)
-    role = serializers.CharField(source="get_role_display")
+    user = WriteableUserEmailSerializer(
+        allow_null=False,
+        slug_field="email",
+        queryset=User.objects.all(),
+        validators=[UniqueForSite(queryset=Membership.objects.all())],
```
sarahfirstvoices, line 72: Should the queryset here be Users instead of Memberships?

firstvoices/backend/views/membership_views.py
```diff
@@ -40,7 +109,28 @@ class MembershipViewSet(
...
+        super().initial(*args, **kwargs)
+
+        if self.request.method.lower() in ["put", "patch", "delete"]:
+            requesting_user = self.request.user
+            membership = Membership.objects.filter(pk=kwargs["pk"]).first()
+
+            # If the membership being changed is that of a Language Admin
+            # ensure that the requesting user is at least staff
+            if membership.role == Role.LANGUAGE_ADMIN:
+                if hasattr(requesting_user, "app_role"):
```
sarahfirstvoices, line 124: A better option here would be to call the permission predicate as a function, so you don't have to worry about checking all the edge cases. (I think this might need additional handling for the AnonymousUser otherwise.) So this could be `if predicates.is_at_least_staff_admin(requesting_user, None)...`

firstvoices/backend/views/membership_views.py
```diff
@@ -30,6 +34,71 @@
             403: OpenApiResponse(description=doc_strings.error_403),
             404: OpenApiResponse(description=doc_strings.error_404),
         },
+        parameters=[
+            site_slug_parameter,
+            id_parameter,
+        ],
+    ),
+    create=extend_schema(
+        description=_("Add a membership. User cannot be changed after creation."),
```
sarahfirstvoices, line 43: +1 for the detailed doc strings here

firstvoices/backend/serializers/membership_serializers.py
```diff
@@ -54,19 +56,37 @@ class Meta:
...
+    )
+    role = WritableRoleField(required=True)
+
+    def update(self, instance, validated_data):
+        """
+        Override update to make user read only after creation.
+        """
+        validated_data.pop("user", None)
+        return super().update(instance, validated_data)
```
sarahfirstvoices, line 81: This looks good

sarahfirstvoices (changes requested):

firstvoices/backend/serializers/fields.py
gmcauliffe: Actually, aren't we using the WritableVisibilityField for Visibility? (directly above)

sarahfirstvoices: Oh dear, looks like we are using a mix of both. I was looking at this spot where we use EnumField: https://github.com/First-Peoples-Cultural-Council/fv-be/blob/87861203fd0fb8be538aaea5e386cdccdd0edcef/firstvoices/backend/serializers/job_serializers.py#L96 

If the generic EnumField works for this case, let's use it, and at some point work on moving the rest of the visibility fields to use it as well. I don't think the bespoke versions are doing anything different, but correct me if I'm missing something

firstvoices/backend/serializers/membership_serializers.py
sarahfirstvoices: Gotcha gotcha. Looking more closely, this validator only works for site content models so I think this is the best option. Thanks for checking!

firstvoices/backend/serializers/fields.py
gmcauliffe: The EnumField does a couple of things differently. The first being that It doesn't allow for labels with spaces (such as "Language Admin") which is fairly easily accommodated without affecting other uses of the field by adding `.replace(" ", "_")`. The second difference is it strips casing from the labels it returns (`.lower()`) and many other tests for other apis are expecting it to do this. Membership is currently cased both in the fe and in `get_role_display`.

PR #1376, opened 2025-07-16
FW-6648 Add subtring matching to search query builder

sahil97:
### Description of Changes
- Added substring matching to search query builder
- Updated search query builder to use reusable methods
- Added integration tests

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/.coveragerc
+++ b/firstvoices/.coveragerc
@@ -3,6 +3,7 @@ source=.
 branch=True
 omit =
     */tests/*
+    */integration_tests/*
     */migrations/*
     */backend/admin/*
     */__init__.py
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,6 +35,6 @@ ignore_errors = True
 
 [coverage:run]
 include = firstvoices/**
-omit = */migrations/*, */tests/*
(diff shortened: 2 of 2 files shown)


bronson-c (approved):
This looks good! My one concern from this PR is that prefix substring matches will have the same boost as middle or suffix substrings. For example if "blowfish" and "fisherman" were both in the dictionary and "fish" was queried, the order may be:

- blowfish
- fisherman

I'll make this a separate ticket, but I think the order should be 
1. Exact match
2. prefix substring
3. middle-fix = suffix
4. fuzzy

gmcauliffe:
> This looks good! My one concern from this PR is that prefix substring matches will have the same boost as middle or suffix substrings. For example if "blowfish" and "fisherman" were both in the dictionary and "fish" was queried, the order may be:
> 
>     * blowfish
> 
>     * fisherman
> 
> 
> I'll make this a separate ticket, but I think the order should be
> 
>     1. Exact match
> 
>     2. prefix substring
> 
>     3. middle-fix = suffix
> 
>     4. fuzzy

Great idea. Before the next change we should also chat with James and get a linguist's perspective. Prefix substrings would definitely take priority in English, but he might have more information about how that applies to other languages.

PR #1438, opened 2025-09-23
FW-6776 Make media validation errors in batches more verbose

sahil97:
### Description of Changes
- Updated error messages for batch validation
- Lowercased all columns before they go in for media validation

### Checklist

### Additional Notes
N/A

### DIFF
--- a/firstvoices/backend/views/import_job_views.py
+++ b/firstvoices/backend/views/import_job_views.py
@@ -186,7 +186,6 @@ class ImportJobViewSet(
                 ignore_result=True,
             )
         )
-        # validate_import_job(str(import_job_id))
 
         return Response(status=status.HTTP_202_ACCEPTED)
 


bronson-c (approved):
LGTM! Just for clarification, now all columns out of a batch import job will be lower-cased, but no need to tell users to lower case their column titles yes?

gmcauliffe:
> > LGTM! Just for clarification, now all columns out of a batch import job will be lower-cased, but no need to tell users to lower case their column titles yes?
> 
> Yes, the columns should act as case-insensitive, and will be lower-cased while processing, which was their behaviour initially for dictionary entries, now they work similarly for media as well.

Do we have a test for the case insensitivity of column titles? If not, it might be worth adding a dedicated test for that if we are removing the variety of casing from other tests.

PR #1679, opened 2026-05-28
FW-7034: Add notify endpoint for import-jobs

sahil97:
### Description of Changes
- Added notify endpoint, added new enum for ImportJobStatus and a new status

### Checklist

### Additional Notes
N/A


firstvoices/backend/tasks/update_job_tasks.py
```diff
@@ -178,10 +178,10 @@ def run_update_job(data, update_job):
             missing_referenced_media,
             dry_run=False,
         )
-        update_job.status = JobStatus.COMPLETE
+        update_job.status = ImportJobStatus.COMPLETE
```
gmcauliffe, line 181: Should update jobs have their own statuses? If not, should we change "ready_for_import" to "ready_for_confirmation" or something else that could apply to updates as well?

gmcauliffe (approved):
LGTM. Just one question about Import jobs vs Update Jobs, but potentially that will be addressed in the upcoming separation of the two.

firstvoices/backend/tasks/update_job_tasks.py
bronson-c: When import jobs and update jobs are separated they will get their own sets of statuses, so I think this can remain "ready_for_import", @sahil97 thoughts?

bronson-c (approved):
LGTM!

firstvoices/backend/tasks/update_job_tasks.py
sahil97: Right, the plan is to have separate status enum for Update jobs as well, which will be done when we separate them from the import-jobs. So, "ready_for_import" should work here.

